PR #18, opened 2021-12-16
Adjusted networking suite to the new tnf design

kononovn:
Renamed default network test case
Removed partner pod support
Refactored netgative disconnected test case


theemj (approved):
lgtm!

PR #20, opened 2021-12-23
Automate remaining container cert tests

theemj:


tests/globalhelper/helper.go
```diff
@@ -69,6 +69,12 @@ func defineCertifiedContainersInfo(config *globalparameters.TnfConfig, certified
 
 	for _, certifiedContainerFields := range certifiedContainerInfo {
 		nameRepository := strings.Split(certifiedContainerFields, "/")
+		if len(nameRepository) == 1 {
+			// certifiedContainerInfo item does not contain separation character
+			// use this to add only the Certifiedcontainerinfo field with no sub fields
+			config.Certifiedcontainerinfo = append(config.Certifiedcontainerinfo)
```
theemj, line 75: fixing this now

tests/affiliatedcertification/affiliatedcertparameters/paramaters.go
```diff
@@ -8,4 +8,8 @@ var (
 	CertifiedContainerNodeJsUbi          = "nodejs-12/ubi8"
 	CertifiedContainerRhel7OpenJdk       = "openjdk-11-rhel7/openjdk"
 	UncertifiedContainerFooBar           = "foo/bar"
+	ContainerFieldNoSubFields            = ""
```
kononovn, line 11: Please remove this var. It used only once

tests/affiliatedcertification/tests/affiliated_certification_container.go
```diff
@@ -121,33 +121,148 @@ var _ = Describe("Affiliated-certification container certification,", func() {
 
 	// 46566
 	It("certifiedcontainerinfo field exists in tnf_config but has no value [skip]", func() {
-		Skip("Under development")
+		By("Add container information to tnf_config.yml")
```
kononovn, line 124: Please change to By("Add container information to " + globalparameters.DefaultTnfConfigFileName)

tests/affiliatedcertification/tests/affiliated_certification_container.go
```diff
@@ -121,33 +121,148 @@ var _ = Describe("Affiliated-certification container certification,", func() {
...
+		err = nethelper.ValidateIfReportsAreValid(
+			affiliatedcertparameters.TestCaseContainerAffiliatedCertName,
+			globalparameters.TestCaseSkipped)
+		Expect(err).ToNot(HaveOccurred(), "Error validating test reports")
 	})
 
 	// 46567
 	It("certifiedcontainerinfo field does not exist in tnf_config [skip]", func() {
-		Skip("Under development")
+		By("Add container information to tnf_config.yml")
```
kononovn, line 148: It looks like all tests have the same flow. May be we can move the main test flow in to helper function and call it from test cases. What do you think?

kononovn (changes requested):
Looks good, just couple of more minor changes

tests/affiliatedcertification/tests/affiliated_certification_container.go
theemj: makes sense to me - will do

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	By("Verify test case status in Junit and Claim reports")
+
+	err = nethelper.ValidateIfReportsAreValid(
+		affiliatedcertparameters.TestCaseContainerAffiliatedCertName,
+		expectedResult)
+	Expect(err).ToNot(HaveOccurred(), "Error validating test reports")
+
+	if err != nil {
```
kononovn, line 48: Considering expression above this condition is useless. Let remove all ginkgo calls from affiliatedcerthelper package(Expect) except By("") and return just clear error which will be processes in a test case later.

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -0,0 +1,61 @@
+package affiliatedcerthelper
+
+import (
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
```
kononovn, line 7: Please don't use anonymous import in affiliatedcerthelper module

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	ginkgo.By("Start test")
+
+	err = globalhelper.LaunchTests(
+		[]string{affiliatedcertparameters.AffiliatedCertificationTestSuiteName},
+		affiliatedcertparameters.TestCaseContainerSkipRegEx,
+	)
+
+	if strings.Compare(expectedResult, globalparameters.TestCaseFailed) == 0 {
```
kononovn, line 36: You can simplify this condition by:

if strings.Contains(expectedResult, globalparameters.TestCaseFailed) && err == nil {
    ....
}

if strings.Contains(expectedResult, globalparameters.TestCasePassed) && err != nil {
    ....
}

kononovn (approved):
Look good to me! Thanks

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -0,0 +1,58 @@
...
+		[]string{affiliatedcertparameters.AffiliatedCertificationTestSuiteName},
+		affiliatedcertparameters.TestCaseContainerSkipRegEx,
+	)
+
+	if strings.Contains(expectedResult, globalparameters.TestCaseFailed) && err == nil {
+		return fmt.Errorf("error running %s test",
+			affiliatedcertparameters.AffiliatedCertificationTestSuiteName)
+	}
+
+	if (strings.Contains(expectedResult, globalparameters.TestCasePassed) ||
```
edcdavid, line 41: No sure I understand the reason for this test and the one above. Shouldn't err simply be always nil, irrespective of the expected result? if the exec cmd fails, success and failure results should be equally not valid. Maybe I am missing something else?

sebrandon1 (approved):

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
theemj: The first case (TestCaseFailed and error is nil) if for negative tests: our expected outcome is an error, so the case where an error isn't present is the one we need to throw an error about in the testing.

edcdavid: I see, thanks. The exit code is actually non zero in case test cases fail

edcdavid (approved):

PR #25, opened 2022-01-12
Automate two lifecycle-shutdown preStop flag tests

shaior:
47311, 47315 lifecycle shutdown tests

### DIFF
--- a/tests/lifecycle/lifecycle_suite_test.go
+++ b/tests/lifecycle/lifecycle_suite_test.go
@@ -2,7 +2,6 @@ package lifecycle
 
 import (
 	"flag"
-	"fmt"
 	"runtime"
 	"testing"
 	"time"
@@ -62,15 +61,15 @@ var _ = BeforeSuite(func() {
 
 var _ = AfterSuite(func() {
 
-	By(fmt.Sprintf("Remove %s namespace", lifeparameters.LifecycleNamespace))
-	err := namespaces.DeleteAndWait(
-		globalhelper.APIClient,
-		lifeparameters.LifecycleNamespace,
-		lifeparameters.WaitingTime,
-	)
-	Expect(err).ToNot(HaveOccurred())
+	// 	By(fmt.Sprintf("Remove %s namespace", lifeparameters.LifecycleNamespace))
+	// 	err := namespaces.DeleteAndWait(
+	// 		globalhelper.APIClient,
+	// 		lifeparameters.LifecycleNamespace,
+	// 		lifeparameters.WaitingTime,
+	// 	)
+	// 	Expect(err).ToNot(HaveOccurred())
 
-	By("Remove reports from reports directory")
-	err = globalhelper.RemoveContentsFromReportDir()
-	Expect(err).ToNot(HaveOccurred())
+	// 	By("Remove reports from reports directory")
+	// 	err = globalhelper.RemoveContentsFromReportDir()
+	// 	Expect(err).ToNot(HaveOccurred())
 })
--- a/tests/lifecycle/lifecycle_suite_test.go
+++ b/tests/lifecycle/lifecycle_suite_test.go
@@ -2,6 +2,7 @@ package lifecycle
 
 import (
 	"flag"
+	"fmt"
 	"runtime"
 	"testing"
 	"time"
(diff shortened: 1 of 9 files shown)


tests/lifecycle/lifecycle_suite_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"flag"
+	"fmt"
+	"runtime"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/reporters"
+
+	lifecycleparameters "github.com/test-network-function/cnfcert-tests-verification/tests/lifecycle/lifeparameters"
```
theemj, line 13: this could be simplified if you change the package name in your parameters.go file to match the folder name (lifeparameters). Or if you wanted it to be called lifecycleparameters you could change the folder name. Either way, you would no longer need to specify the name in the import.

tests/lifecycle/lifecycle_suite_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	By("Create namespace")
+	err = namespaces.Create(lifecycleparameters.LifecycleNamespace, globalhelper.APIClient)
+	Expect(err).ToNot(HaveOccurred())
+
+	By("Define TNF config file")
+	err = globalhelper.DefineTnfConfig(
+		[]string{lifecycleparameters.LifecycleNamespace},
+		[]string{lifecycleparameters.TestPodLabel},
+		[]string{})
```
theemj, line 57: Just a heads-up: now that my changes to DefineTnfConfig for the operator-certification tests have been merged in (https://github.com/test-network-function/cnfcert-tests-verification/pull/24/files#diff-e1c500cd9f2c78d57abb90392858cb8048992645b2eb5693eb098951dff49ada), another empty string array arg will need to be added when you merge the main branch into this one.

theemj (approved):
Looks good! Just one suggestion about simplifying the parameters import and an FYI about a recent change to DefineTnfConfig.

tests/lifecycle/tests/lifecycle_shutdown.go
```diff
@@ -0,0 +1,72 @@
...
+			lifecycleparameters.LifecycleNamespace,
+			globalhelper.Configuration.General.TestImage,
+			lifecycleparameters.TestDeploymentLabels)
+
+		By("Start lifecycle lifecycle-container-shutdown test")
+		err := globalhelper.LaunchTests(
+			[]string{lifecycleparameters.LifecycleTestSuiteName},
+			lifecycleparameters.SkipAllButShutdownRegex,
+		)
+		Expect(err).To(HaveOccurred())
```
greyerof, line 62: Sorry, I think I missed this from previous reviews, but, there's a couple more reasons in the LaunchTests() function for the "err" to have occurred, so you should find a way here to make sure it comes from failing TC/s and not because of the other reasons. I just recommend to modify the LaunchTests function to return both the TCs result (bool) and the err. Anyway, this would be a core change that affects to other test suites, so it could be done in a different PR.

tests/lifecycle/lifecycle_suite_test.go
shaior: Thanks for the heads up :)

tests/lifecycle/tests/lifecycle_shutdown.go
shaior: Thanks for the comment, we will discuss how should we proceed.

tests/lifecycle/lifeparameters/parameters.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+const (
+	WaitingTime   = 5 * time.Minute
+	RetryInterval = 5
+)
+
+var (
+	LifecycleNamespace     = "lifecycle-tests"
+	testPodLabelPrefixName = "test-network-function.com/generic"
```
kononovn, line 15: I think it would be better to change it to "test-network-function.com/lifecycle". What do you think?

tests/lifecycle/lifeparameters/parameters.go
```diff
@@ -0,0 +1,27 @@
...
+	testPodLabelValue      = "target"
```
kononovn, line 16: Can we change it to "testing" ?

tests/lifecycle/tests/lifecycle_shutdown.go
```diff
@@ -0,0 +1,74 @@
...
+		By("Clean namespace before each test")
+		err := namespaces.Clean(lifeparameters.LifecycleNamespace, globalhelper.APIClient)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	// 47311
+	It("One deployment, one pod with one container that has preStop field configured", func() {
+
+		By("Define deployment with preStop field configured")
+		preStopCommand := []string{"/bin/sh", "-c", "killall -0 tail"}
```
kononovn, line 26: We use this var only in 1 place. Please remove this var and paste this command directly into func's argument field

tests/lifecycle/tests/lifecycle_shutdown.go
```diff
@@ -0,0 +1,74 @@
...
+		preStopDeploymentStruct := deployment.RedefineWithPreStopSpec(deployment.DefineDeployment(
+			lifeparameters.LifecycleNamespace,
+			globalhelper.Configuration.General.TnfImage,
+			lifeparameters.TestDeploymentLabels), preStopCommand)
+		err := nethelper.CreateAndWaitUntilDeploymentIsReady(preStopDeploymentStruct, lifeparameters.WaitingTime)
```
kononovn, line 31: Code duplication between 2 test cases. Can you move common code in to a helper function ?

kononovn (changes requested):

greyerof (approved):

tests/utils/deployment/deployment.go
```diff
@@ -1,17 +1,69 @@
...
+	"github.com/test-network-function/cnfcert-tests-verification/tests/globalhelper"
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
+
+	. "github.com/onsi/gomega"
 )
 
+func isDeploymentReady(operatorNamespace string, deploymentName string) (bool, error) {
```
kononovn, line 18: In Utils we keep only define functions. Please move all non-define function to  ./tests/globalhelper/deployment.go

kononovn (changes requested):
Looks good, just 1 more step

kononovn (approved):
Looks good to me

PR #31, opened 2022-01-26
Bump GolangCI-lint to v1.44.0

sebrandon1:
https://github.com/golangci/golangci-lint/releases/tag/v1.44.0

```
INFO [runner] linters took 10.659234125s with stages: goanalysis_metalinter: 10.652175941s 
tests/utils/config/config.go:41:6: variable name 'c' is too short for the scope of its usage (varnamelen)
        var c Config
```


theemj (approved):
looks good and works great!

PR #33, opened 2022-02-02
Automated lifecycle-terminationGracePeriod tests

shaior:


kononovn (approved):
Looks good to me +1

tests/lifecycle/tests/lifecycle_termination_grace_period.go
```diff
@@ -0,0 +1,107 @@
...
+		By("Start lifecycle lifecycle-pod-termination-grace-period test")
+		err = globalhelper.LaunchTests(
+			[]string{lifeparameters.LifecycleTestSuiteName},
+			lifeparameters.SkipAllButTerminationGracePeriodRegex)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Verify test case status in Junit and Claim reports")
+		err = globalhelper.ValidateIfReportsAreValid(
+			lifeparameters.TerminationGracePeriodDefaultName,
+			globalparameters.TestCasePassed)
```
greyerof, line 44: This TC passes on 3.2 but will fail on 3.3 when it's released because of this PR https://github.com/test-network-function/test-network-function/pull/562
With that PR, we can detect whether the terminationGracePeriod value was actually set by the user (no matter which value), which was the original aim of this TNF test case.

tests/lifecycle/tests/lifecycle_termination_grace_period.go
```diff
@@ -0,0 +1,107 @@
...
```
greyerof, line 104: Same here.

greyerof (approved):

PR #36, opened 2022-02-07
Added 5 new test cases to networking multus suite

kononovn:
Added TC Networking custom namespace, custom daemonset 3 pods with skip label [skip]
Added TC Networking custom namespace, custom deployment and daemonset 3 pods with skip label[skip]
Added TC Networking custom namespace, custom deployment and daemonSet 3 pods, daemonSet has skip label
Added TC Networking custom namespace, custom deployment 3 pods, 2 NADs, multiple Multus interfaces on deployment
Added TC Networking custom namespace, custom deployment and daemonset 3 pods, 2 NADs, multiple Multus interfaces on deployment no connectivity via secondary interface[negative]
Fixed Multus test cases names

### DIFF
--- a/tests/networking/tests/networking_multus_links_negative.go
+++ b/tests/networking/tests/networking_multus_links_negative.go
@@ -68,7 +68,7 @@ var _ = Describe("Networking custom namespace, ", func() {
 
 			By("Define deployment and create it on cluster")
 			err = nethelper.DefineAndCreateDeploymentWithMultusOnCluster(
-				netparameters.TestDeploymentAName, netparameters.TestNadNameA, 3)
+				netparameters.TestDeploymentAName, []string{netparameters.TestNadNameA}, 3)
 			Expect(err).ToNot(HaveOccurred())
 
 			By("Start tests")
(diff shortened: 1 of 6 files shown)


tests/networking/tests/networking_multus_links_negative.go
```diff
@@ -124,4 +124,43 @@ var _ = Describe("Networking custom namespace, ", func() {
 				globalparameters.TestCaseFailed)
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+	It("custom deployment and daemonset 3 pods, 2 NADs, multiple Multus interfaces on deployment no "+
+		"connectivity via secondary interface[negative]", func() {
+
+		if len(multusInterfaces) < 2 {
```
hamadise, line 131: this is where I have some concerns, should not you let the test run and found that it's being skipped ?

kononovn: > this is where I have some concerns, should not you let the test run and found that it's being skipped ?

Yes. It's Ok. I still want to run all possible test cases on cluster that doesn't have 2 secondary interfaces

theemj (approved):

PR #39, opened 2022-02-15
Added affiliated-certification namespace and pod label

theemj:
Replaced netparameters namespace and pod label in affiliated certification suite with affiliated certification-specific namespace and labels.


kononovn (changes requested):
Looks good, just 2 small comments

PR #40, opened 2022-02-15
Add and Delete label func to affiliatedcerthelper

kononovn:


theemj (approved):
looks good to me :+1:

PR #42, opened 2022-02-17
Fixed networking multus flaky test cases

kononovn:
Removed networking_multus_links_negative.go file
Removed PutDownInterfaceOnNode and PutAllInterfaceUPonAllNodes from
nethelper


tests/networking/tests/networking_multus_links.go
```diff
@@ -373,4 +373,123 @@ var _ = Describe("Networking custom namespace, ", func() {
 		Expect(err).ToNot(HaveOccurred())
 
 	})
+
```
edcdavid, line 376: // 48346

tests/networking/nethelper/nethelper.go
```diff
@@ -370,7 +370,11 @@ func getInterfacesList(runningPod corev1.Pod) ([]string, error) {
 	}
 
 	for _, nodeInterface := range linuxInterfaces {
-		if nodeInterface.Master == "" && nodeInterface.Operstate == "UP" {
+		if nodeInterface.Master == "" &&
+			!strings.Contains(nodeInterface.IfName, "ovn") &&
```
edcdavid, line 374: This fixes the issue where multus if could be created on a non external if, so if shutdown traffic still up. But now I guess even not fixing this since we turn the port off in the pod it would still work

edcdavid (approved):

theemj (approved):

tests/networking/nethelper/nethelper.go
kononovn: yes, right

PR #44, opened 2022-02-24
Automated 5 pod-high-availability tests

shaior:


greyerof (approved):

theemj (approved):

tests/lifecycle/lifeparameters/parameters.go
```diff
@@ -23,27 +23,33 @@ var (
 	LifecycleTestSuiteName  = "lifecycle"
 	SkipAllButShutdownRegex = "lifecycle-pod-high-availability lifecycle-pod-scheduling" +
```
kononovn, line 24: I think it would be better if can create 1 list with all name of test cases and then remove only relevant test case name from it.

kononovn (changes requested):

tests/lifecycle/lifeparameters/parameters.go
shaior: good idea,
I will check it after the pod-recreation issue is solved.

shaior: as agreed, this will be done in the next PR.

kononovn (approved):

PR #49, opened 2022-03-10
Add tnf trace reports for each test

shaior:
A folder would be generated for each test case containing reports that would be generated for each of the specs.


tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -33,6 +33,8 @@ func SetUpAndRunContainerCertTest(containersInfo []string, expectedResult string
 
 	err = globalhelper.LaunchTests(
 		[]string{affiliatedcertparameters.AffiliatedCertificationTestSuiteName},
+		"Add container information to "+globalparameters.DefaultTnfConfigFileName,
```
theemj, line 36:
```suggestion
		affiliatedcertparameters.TestCaseContainerAffiliatedCertName,
```

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -76,6 +78,8 @@ func SetUpAndRunOperatorCertTest(operatorsInfo []string, expectedResult string)
...
```
theemj, line 81:
```suggestion
		affiliatedcertparameters.TestCaseOperatorAffiliatedCertName,
```

theemj (approved):

kononovn (changes requested):
@shaior Please try to use builtin  function ginkgo.CurrentGinkgoTestDescription().FullTestText :) . BTW this function has a lot of useful information CurrentGinkgoTestDescription

tests/globalhelper/runhelper.go
```diff
@@ -47,5 +48,44 @@ func LaunchTests(testSuites []string, skipRegEx string) error {
 	cmd.Args = append(cmd.Args, testArgs...)
 	cmd.Dir = Configuration.General.TnfRepoPath
 
+	if os.Getenv("DEBUG_TNF") == "true" {
```
kononovn, line 51: Please initiate DEBUG_TNF and LOG_LEVEL in config.go (Config struct) with rest of env vars.

tests/globalhelper/runhelper.go
```diff
@@ -47,5 +48,44 @@ func LaunchTests(testSuites []string, skipRegEx string) error {
...
+		os.Setenv("LOG_LEVEL", "trace")
+
+		buildNumber, present := os.LookupEnv("BUILD_NUMBER")
```
kononovn, line 54: What is BUILD_NUMBER and why do we need it?

tests/globalhelper/runhelper.go
```diff
@@ -47,5 +48,44 @@ func LaunchTests(testSuites []string, skipRegEx string) error {
...
+
+		var debugFolderName string
+
+		if present {
+			debugFolderName = "Debug-" + buildNumber
```
kononovn, line 59: No need for build number. Let's just overwrite debugs after every run.

tests/globalhelper/runhelper.go
```diff
@@ -47,5 +48,44 @@ func LaunchTests(testSuites []string, skipRegEx string) error {
...
+		} else {
+			debugFolderName = "Debug-Manual"
+
+		}
+
+		folderPath := filepath.Join(Configuration.General.ReportDirAbsPath, debugFolderName, tcNameForFolder)
```
kononovn, line 65: Let's build following struct Configuration.General.ReportDirAbsPath/Test_suite_name_folder/Test_case_name_folder.

After every run we will overwrite Configuration.General.ReportDirAbsPath/**Test_suite_name_folder**

kononovn (changes requested):
It looks very good. Just move this logic to config.go file

tests/globalhelper/runhelper.go
shaior: This env var will be passed from Jenkins, and its the build number.
We need to be able to distinguish between ci logs, in order to see what went wrong in the specific build.

tests/globalhelper/runhelper.go
shaior: I think we should save those logs for a certain amount of time, so we would be able to
check those logs.
if we will overwrite them, then we lost the logs of the build.

or we are saving them in a different place?

tests/globalhelper/runhelper.go
shaior: Do you mean in the AfterSuite?

again, but then we lost the logs, we need them to see what went ok/ wrong.

greyerof (commented):
I recommend using the following structure for tnf debug log folder+file names:
`tnf_test_suite_name/tnf_test_case_name/qe_name_or_id.log`

Where:
```
tnf_test_suite_name:
	observability
	operator
	accesscontrol
	lifecycle
	...
```

```
tnf_test_case_name:
	pod-recreation
	pod-scheduling
	...
```
For qe_name_or_id, I don't recommend the "It" blocks text, as it's usually too long. I'd use shorter descriptive names (difficult) or just internal QE TC numbers. You could use polarion ones or create new ones just for this debug logging files purpose. I'd go with polarion numbres.

tests/globalhelper/helper.go
```diff
@@ -264,3 +265,16 @@ func AppendContainersToDeployment(deployment *v1.Deployment, containersNum int,
 
 	return deployment
 }
+
+// VerifyFolderExists verifies if a given folder exists, and creates it if it doesn't.
+func VerifyFolderExists(folderPath string, perm fs.FileMode) (bool, error) {
```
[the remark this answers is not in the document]
shaior: thanks

tests/globalhelper/reporthelper.go
```diff
@@ -188,3 +189,10 @@ func removeCharactersFromString(stringToFormat string, charactersToRemove []stri
 func formatTestCaseName(tcName string) string {
 	return removeCharactersFromString(tcName, []string{"-", "_", " "})
 }
+
+func ConvertSpecNameToFileName(stringToFormat string) string {
```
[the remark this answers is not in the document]
shaior: agree

tests/globalhelper/reporthelper.go
```diff
@@ -102,6 +102,7 @@ func RemoveContentsFromReportDir() error {
 	}
 	defer tnfReportDir.Close()
 	names, err := tnfReportDir.Readdirnames(-1)
+	fmt.Println("RemoveContentsFromReportDir names = ", names)
```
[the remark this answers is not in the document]
shaior: this is a leftover from debugging, removed it, thanks for noticing.

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -33,6 +33,8 @@ func SetUpAndRunContainerCertTest(containersInfo []string, expectedResult string
 
 	err = globalhelper.LaunchTests(
 		[]string{affiliatedcertparameters.AffiliatedCertificationTestSuiteName},
+		affiliatedcertparameters.TestCaseContainerAffiliatedCertName,
```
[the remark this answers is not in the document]
shaior: agree,
@theemj  Hey Emily, please let me know how can I make the folder name shorter

tests/globalhelper/runhelper.go
shaior: done, added deletion in run-tests.sh script.

tests/globalhelper/runhelper.go
shaior: done, we need to save the Debug folder to Jenkins artifacts.

tests/globalhelper/runhelper.go
shaior: Deleted it.

tests/globalhelper/runhelper.go
```diff
@@ -47,5 +48,44 @@ func LaunchTests(testSuites []string, skipRegEx string) error {
 	cmd.Args = append(cmd.Args, testArgs...)
 	cmd.Dir = Configuration.General.TnfRepoPath
 
+	if os.Getenv("DEBUG_TNF") == "true" {
+		os.Setenv("LOG_LEVEL", "trace")
```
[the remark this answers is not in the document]
shaior: Done.

tests/globalhelper/runhelper.go
shaior: Done.

scripts/run-tests.sh
```diff
@@ -6,8 +6,13 @@ export ACK_GINKGO_DEPRECATIONS=1.16.5
 export ACK_GINKGO_RC=true
 EXCLUDED_FOLDERS=""
 ALL_TESTS_FOLDERS=$(ls -d ./tests/*/)
+REPORTS_FOLDER="./reports/Debug"
```
kononovn, line 9: Please remove this code

scripts/run-tests.sh
```diff
@@ -6,8 +6,13 @@ export ACK_GINKGO_DEPRECATIONS=1.16.5
...
 
 function run_tests {
+    if [ -d "$REPORTS_FOLDER" ]; then
```
kononovn, line 12: Please remove this code. We should handle it in the project code instead of external bash script

tests/lifecycle/tests/lifecycle_pod_owner_type.go
```diff
@@ -115,8 +123,7 @@ var _ = Describe("lifecycle lifecycle-pod-owner-type", func() {
 	})
 
 	// 47430
-	It("Multiple deployments, 1 pod not related to any resource [negative]", func() {
-
+	It("Two deployments, 1 pod not related to any resource [negative]", func() {
```
kononovn, line 126: Please select either Two or 2 like in the previous test case

tests/utils/config/config.go
```diff
@@ -82,6 +84,50 @@ func NewConfig() (*Config, error) {
...
+	folderPath := filepath.Join(c.General.ReportDirAbsPath, "Debug", testSuite)
+
+	_, err := CreateFolder(folderPath, 0755)
+
+	if err != nil {
+		panic(err)
+	}
+
+	tcFile := filepath.Join(folderPath, tcName+".log")
+	outfile, err := os.OpenFile(tcFile, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0755)
```
kononovn, line 122: You can just overwrite the content of the file  instead of deleting Debug folder in bash script.

kononovn (changes requested):

tests/utils/config/config.go
shaior: it will delete the content of the file in each launchTests func call, which means in each test run, that's not what we want.

tests/lifecycle/tests/lifecycle_pod_owner_type.go
shaior: done, thanks.

scripts/run-tests.sh
shaior: this is the first script that starts running all tests, the problem is that 
we need to delete it only once, at the beginning of the code run, so we cant do it 
in one of the suites.

scripts/run-tests.sh
shaior: done.

scripts/run-tests.sh
shaior: done. handling it in DefineLogFile func

tests/utils/config/config.go
shaior: Done.

tests/globalhelper/runhelper.go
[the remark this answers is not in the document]
greyerof: I believe that's also a bit confusing, Nikita, as usually TNF refers to the test-network-function suite that this QE repo is testing. I know we should change our tnf LOG_LEVEL variable into TNF_LOG_LEVEL, but yours should be something like QE_LOG_LEVEL or TNF_QE_LOG_LEVEL.
Basically, my recommendation is that in your QE repo, TNF/tnf prefix should always refer to the test-network-function test suite names to avoid confusion, and you should use something like QE/qe prefixes to your variables/function when distinguish them from the context is not so straightforward.

tests/utils/config/config.go
```diff
@@ -82,6 +84,60 @@ func NewConfig() (*Config, error) {
...
+	if err != nil {
+		panic(err)
+	}
+
+	tcFile := filepath.Join(folderPath, tcName+".log")
+
+	// if the log file already exists, remove it and create a new one.
+	if _, err := os.Stat(tcFile); err == nil {
+		err = os.Remove(tcFile)
+
```
greyerof, line 126: Recommendation: do not add extra blank lines before the "if err != nil".

tests/utils/config/config.go
```diff
@@ -82,6 +84,60 @@ func NewConfig() (*Config, error) {
...
+	err = os.MkdirAll(folderPath, perm)
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func (c *Config) DefineLogFile(testSuite string, tcName string) *os.File {
```
greyerof, line 112: DefineLogFile -> CreateLogFile

tests/globalhelper/runhelper.go
```diff
@@ -35,17 +35,35 @@ func LaunchTests(testSuites []string, skipRegEx string) error {
 		glog.V(5).Info(fmt.Sprintf("set skip regex to %s", skipRegEx))
 	}
 
-	if len(testSuites) > 0 {
+	if len(testSuite) > 0 {
```
greyerof, line 38: Is there any use for LunchTests where len(testSuite) == 0 ? If there's no such case, you can just do something like:
```
if testSuite == "" {
    return errors.New("no test suite name provided.")
}
```

greyerof (approved):

tests/utils/config/config.go
shaior: Fixed. will upload in the commit. thanks

tests/utils/config/config.go
shaior: Fixed, thanks

tests/globalhelper/runhelper.go
shaior: thanks!

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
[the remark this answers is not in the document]
theemj: hmm we could just use "container-is-certified" and "operator-is-certified", we'd just need to add two parameters to affiliatedcertparameters like:

```
ContainerTestCaseName = "container-is-certified"
OperatorTestCaseName =  "operator-is-certified"
```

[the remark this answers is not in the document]
shaior: @theemj  Let's talk about it please, because only one file is being generated for all of your tests, and it keeps getting deleted and written again every new launch, thus every test needs a different name.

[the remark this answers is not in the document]
theemj: Ah I see. I could change the SetUpAndRunContainerCertTest parameters so that a unique name is passed in for each test case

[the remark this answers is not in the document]
shaior: @theemj Hey, attached a screenshot showing the tree of the file inside affiliated folder, please confirm that this is ok.
also, 3 tests are missing because they have failed before running the tnf, so no log file is generated for them, the log file contains only tnf logs, not QE logs.

image

theemj (approved):
lgtm!

tests/globalhelper/reporthelper.go
```diff
@@ -188,3 +188,12 @@ func removeCharactersFromString(stringToFormat string, charactersToRemove []stri
 func formatTestCaseName(tcName string) string {
 	return removeCharactersFromString(tcName, []string{"-", "_", " "})
 }
+
+func ConvertSpecNameToFileName(specName string) string {
```
kononovn, line 192: Can you rewrite this function using for loop?
for _, symbol := range []string(" ",", ","-") {
....
}

tests/globalhelper/runhelper.go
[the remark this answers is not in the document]
kononovn: Agree. @shaior Please re-name this env var to VERIFICATION_LOG_LEVEL. @greyerof What do you think?

tests/utils/config/config.go
```diff
@@ -82,6 +84,59 @@ func NewConfig() (*Config, error) {
 	return &conf, nil
 }
 
+func (c *Config) DebugTnf() (bool, error) {
```
kononovn, line 87: This function never return error. Please remove error from return elements

tests/utils/config/config.go
```diff
@@ -82,6 +84,59 @@ func NewConfig() (*Config, error) {
...
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// CreateFolder verifies if a given folder exists, and creates it if it doesn't.
+func CreateFolder(folderPath string, perm fs.FileMode) (bool, error) {
+	_, err := os.Stat(folderPath)
+	if !os.IsNotExist(err) {
```
kononovn, line 100: Not sure I understand this condition. Why do we need it ?

tests/utils/config/config.go
```diff
@@ -82,6 +84,59 @@ func NewConfig() (*Config, error) {
...
+		os.Setenv("LOG_LEVEL", "trace")
+
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// CreateFolder verifies if a given folder exists, and creates it if it doesn't.
+func CreateFolder(folderPath string, perm fs.FileMode) (bool, error) {
```
kononovn, line 98: You don't use bool parameter from this func. Please just remove it

kononovn (changes requested):
Just few minor changes are needed.

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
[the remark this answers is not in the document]
kononovn: @greyerof Yes the name of the files are long but on the other hand it's easy to distinguish between test cases. We can not use polarion tags because it's our internal tool.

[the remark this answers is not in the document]
greyerof: I'm ok if it works for you, guys.
I've worked with both methods (numbers and long names) in the past, and just as per my experience, it was actually harder navigating through folders with so long file names inside. It made copy/moving operations more difficult and I ended up renaming the failing TC ones to shorter names when I needed to open them for review.
At the end of the day, the normal scenario is that you'll get only one or two failing TCs/regressions per QE run due someone adding wrong code in the tnf repo. Searching for "23433.log" and "55043.log" files (both example tc numbers) and their code in your repo based with a grep of those numbers seems easier/faster than dealing with those long names.
Also, you don't really need to use polarion numbers. You can start a new numbering scheme independent from polarion's one. For example, first QE TS (networking?) can start with 101, 102, 103... second test suite (lifecycle?) can start with 201, 202, 203... I understand you may not want supporting two different numberings schemes, though.
Anyway, as I said, that's all based on my personal experience, I've never worked on QE, so I only see these things from the developer's side.

tests/globalhelper/reporthelper.go
```diff
@@ -188,3 +188,14 @@ func removeCharactersFromString(stringToFormat string, charactersToRemove []stri
...
+	formatString := specName
+	for _, symbol := range []string{" ", ", ", "-"} {
+		formatString = strings.ReplaceAll(formatString, symbol, "_")
+	}
+
+	formatString = removeCharactersFromString(formatString, []string{","})
```
kononovn, line 198: You can remove it and just `return strings.ToLower(removeCharactersFromString(formatString, []string{","}))`

kononovn (approved):
LGTM

PR #50, opened 2022-03-13
Improved tc naming, improved reportHelper, renamed var on lifeHelper

shaior:


tests/lifecycle/tests/lifecycle_pod_high_availability.go
```diff
@@ -47,12 +47,12 @@ var _ = Describe("lifecycle lifecycle-pod-high-availability", func() {
 		By("Start lifecycle lifecycle-high-availability test")
 		err = globalhelper.LaunchTests(
 			[]string{lifeparameters.LifecycleTestSuiteName},
-			lifeparameters.SkipAllButPodHighAvailabilityRegex)
+			globalhelper.GetStringOfSkipTcs(lifeparameters.SkipsTcsSlice, lifeparameters.PodHighAvailabilityName))
```
theemj, line 50: might be cleaner to call this once and assign it to a variable since it's called multiple times with the same params

theemj (commented):
could call globalhelper.GetStringOfSkipTcs once per test file and assign the value to a variable, otherwise lgtm!

tests/lifecycle/tests/lifecycle_pod_high_availability.go
shaior: good comment, will upload a better version soon, thanks!

tests/globalhelper/reporthelper.go
```diff
@@ -141,7 +141,7 @@ func isTestCaseInExpectedStatusInClaimReport(
 	claimReport claim.Root,
 	expectedStatus string) (bool, error) {
 	for testCaseClaim := range claimReport.Claim.Results {
-		if formatTestCaseName(testCaseName) == formatTestCaseName(testCaseClaim) {
+		if strings.Contains(formatTestCaseName(testCaseClaim), formatTestCaseName(testCaseName)) {
```
greyerof, line 144: Not sure whether this will work if there are tcs that use the name of another tc as prefix. I don't think we have anything like that, but you should protect agains it.

tests/lifecycle/lifeparameters/parameters.go
```diff
@@ -19,37 +19,19 @@ var (
...
-	DeploymentScalingDefaultName      = "lifecycle lifecycle-deployment-scaling"
-	TerminationGracePeriodDefaultName = "lifecycle lifecycle-pod-termination-grace-period"
-	PodOwnerTypeDefaultName           = "lifecycle lifecycle-pod-owner-type"
-	PodRecreationDefaultName          = "lifecycle lifecycle-pod-recreation"
-	PodHighAvailabilityDefaultName    = "lifecycle CNF is designed in high availability mode" +
-		"lifecycle-pod-high-availability"
+	PreStopCommand         = []string{"/bin/sh", "-c", "killall -0 tail"}
+	LifecycleTestSuiteName = "lifecycle"
+
+	SkipsTcsSlice = []string{"lifecycle-pod-recreation", "lifecycle-pod-scheduling",
```
greyerof, line 25: You don't need the "Slice" suffix. Also, you're using it as a skip tc list, but this is just the whole list of TCs in the tnf test suite, so, to me, a better name for this name would be just "lifecycleTnfTestCases", or just "TnfTestCases".

tests/lifecycle/lifeparameters/parameters.go
```diff
@@ -19,37 +19,19 @@ var (
...
+		"lifecycle-deployment-scaling", "lifecycle-pod-termination-grace-period",
+		"lifecycle-pod-owner-type", "lifecycle-container-shutdown", "lifecycle-image-pull-policy",
+		"lifecycle-liveness", "lifecycle-readiness", "lifecycle-statefulset-scaling",
+		"lifecycle-pod-high-availability"}
+
+	ShutdownName               = "lifecycle-container-shutdown"
```
greyerof, line 31: Minor suggestion: add "Tc" and "Tnf" words, to make it more explicit that this vars refer to tnf test cases names.
```
TnfShutdownTcName = "lifecycle-container-shutdown"
TnfDeploymentScalingTcName = ...
...
```

tests/globalhelper/reporthelper.go
shaior: I think that in that case, the tc will have a more unique name that will differentiate it from the other one.
the issue is sometimes tcs on the claim file have a much different name than the actual tc name.

tests/lifecycle/lifeparameters/parameters.go
shaior: Much better name, thank you, renaming it.

tests/lifecycle/lifeparameters/parameters.go
shaior: I agree, renaming, thanks.

greyerof (approved):

theemj (approved):

tests/lifecycle/lifeparameters/parameters.go
```diff
@@ -19,37 +19,19 @@ var (
...
+	PreStopCommand         = []string{"/bin/sh", "-c", "killall -0 tail"}
+	LifecycleTestSuiteName = "lifecycle"
+
+	TnfTestCases = []string{"lifecycle-pod-recreation", "lifecycle-pod-scheduling",
+		"lifecycle-deployment-scaling", "lifecycle-pod-termination-grace-period",
+		"lifecycle-pod-owner-type", "lifecycle-container-shutdown", "lifecycle-image-pull-policy",
+		"lifecycle-liveness", "lifecycle-readiness", "lifecycle-statefulset-scaling",
+		"lifecycle-pod-high-availability"}
+
+	TnfShutdownTcName               = "lifecycle-container-shutdown"
```
kononovn, line 31: Please move test case name vars above TnfTestCases list and re-use them in the list TnfTestCases

tests/globalhelper/reporthelper.go
kononovn: TC names should have full match between junit and claim file. @greyerof Correct me if I'm wrong

tests/lifecycle/lifeparameters/parameters.go
shaior: Good comment, fixing it, thanks !

kononovn (approved):
Looks good to me. Thanks

PR #51, opened 2022-03-13
Added node uncordoning verification to IsClusterStable for better stability

shaior:
Added node uncordoning verification to IsClusterStable for better test cases stability.

### DIFF
--- a/tests/utils/cluster/cluster.go
+++ b/tests/utils/cluster/cluster.go
@@ -9,16 +9,24 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	testclient "github.com/test-network-function/cnfcert-tests-verification/tests/utils/client"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/nodes"
 )
 
 func IsClusterStable(clients *testclient.ClientSet) (bool, error) {
-	nodes, err := clients.Nodes().List(context.Background(), metav1.ListOptions{})
+	nodeInCluster, err := clients.Nodes().List(context.Background(), metav1.ListOptions{})
 	if err != nil {
 		return false, err
 	}
 
-	for _, node := range nodes.Items {
+	for _, node := range nodeInCluster.Items {
 		if node.Spec.Unschedulable {
+			fmt.Println("node", node.Name, "is cordoned, uncordoning it")
+			_, err := nodes.CordonOrUnCordonNodeByName(node.Name, clients, false)
+
+			if err != nil {
+				return false, err
+			}
+
 			glog.V(5).Info(fmt.Sprintf("node %s is in unschedulable state", node.Name))
 
 			return false, nil
--- a/tests/utils/nodes/nodes.go
+++ b/tests/utils/nodes/nodes.go
@@ -2,14 +2,22 @@ package nodes
 
 import (
 	"context"
+	"encoding/json"
 	"time"
 
 	"github.com/golang/glog"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/client"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
+
+type patchStringValue struct {
+	Operation string `json:"op"`
+	Path      string `json:"path"`
+	Value     bool   `json:"value"`
+}
 
 // WaitForNodesReady waits for all nodes become ready.
 func WaitForNodesReady(cs *client.ClientSet, timeout, interval time.Duration) error {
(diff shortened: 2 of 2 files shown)


tests/utils/cluster/cluster.go
```diff
@@ -9,16 +9,24 @@ import (
...
+	nodeInCluster, err := clients.Nodes().List(context.Background(), metav1.ListOptions{})
 	if err != nil {
 		return false, err
 	}
 
-	for _, node := range nodes.Items {
+	for _, node := range nodeInCluster.Items {
 		if node.Spec.Unschedulable {
+			fmt.Println("node", node.Name, "is cordoned, uncordoning it")
+			_, err := nodes.CordonOrUnCordonNodeByName(node.Name, clients, false)
```
greyerof, line 24: You don't need to do this, as it is just a personal preference: I'd create separate functions for each task: nodes.Cordon(name string) & node.Uncordon(name string). There's also no need to include neither the word "Node" nor "Name" in the function name, as it's obvious from the package "nodes" and the argument (name string).

tests/utils/nodes/nodes.go
```diff
@@ -57,3 +65,26 @@ func GetNumOfReadyNodesInCluster(cs *client.ClientSet) (int32, error) {
 
 	return int32(numOfNodesExistsInCluster), nil
 }
+
+// CordonOrUnCordonNodeByName cordones/uncordones a node by a given node name.
+func CordonOrUnCordonNodeByName(nodeName string, clients *client.ClientSet, cordon bool) (bool, error) {
```
greyerof, line 70: As per my previous recommendation (no need to do it): You can have this code inside an internal function in this package called `setUnschedulableValue(clients *client.ClientSet, nodeName string, value bool) (bool, error)`, so it could be called from the exported `nodes.Cordon(string nodeName) (bool, error)` & `nodes.Uncordon(string nodeName) (bool, error)` ones.

tests/utils/cluster/cluster.go
```diff
@@ -9,16 +9,24 @@ import (
...
 
 func IsClusterStable(clients *testclient.ClientSet) (bool, error) {
-	nodes, err := clients.Nodes().List(context.Background(), metav1.ListOptions{})
+	nodeInCluster, err := clients.Nodes().List(context.Background(), metav1.ListOptions{})
 	if err != nil {
 		return false, err
 	}
 
-	for _, node := range nodes.Items {
+	for _, node := range nodeInCluster.Items {
```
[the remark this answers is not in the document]
shaior: correct, it's a typo, fixed.
thanks!

tests/lifecycle/tests/lifecycle_deployment_scaling.go
```diff
@@ -15,8 +15,11 @@ import (
 var _ = Describe("lifecycle lifecycle-deployment-scaling", func() {
 
 	BeforeEach(func() {
+		_, err := lifehelper.ValidateClusterIsStable()
```
greyerof, line 18: So it doesn't make sense that ValidateClusterIsStable returns anything else than the err value. Also, another minor suggestion: a better name for that function could be "WaitUntilClusterIsStable".

greyerof (approved):
Minor suggestions.

tests/utils/cluster/cluster.go
```diff
@@ -19,9 +20,14 @@ func IsClusterStable(clients *testclient.ClientSet) (bool, error) {
 
 	for _, node := range nodes.Items {
 		if node.Spec.Unschedulable {
-			glog.V(5).Info(fmt.Sprintf("node %s is in unschedulable state", node.Name))
+			fmt.Println("node", node.Name, "is cordoned, uncordoning it")
```
kononovn: Please don't use fmt package. Please use glog instead

greyerof (approved):

tests/utils/nodes/nodes.go
```diff
@@ -2,19 +2,27 @@ package nodes
...
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/client"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// WaitForNodesReady waits for all nodes become ready.
-func WaitForNodesReady(cs *client.ClientSet, timeout, interval time.Duration) error {
+type patchStringValue struct {
```
kononovn, line 16: The name of this struct doesn't make sense. Struct can't be an action. Can you please rename ?

tests/utils/cluster/cluster.go
shaior: sure, done.

tests/utils/nodes/nodes.go
```diff
@@ -57,3 +65,26 @@ func GetNumOfReadyNodesInCluster(cs *client.ClientSet) (int32, error) {
...
 	return int32(numOfNodesExistsInCluster), nil
 }
+
+func UnCordon(clients *client.ClientSet, nodeName string) error {
+	return setUnSchedulableValue(clients, nodeName, false)
+}
+
+// setUnSchedulableValue cordones/uncordones a node by a given node name.
+func setUnSchedulableValue(clients *client.ClientSet, nodeName string, unSchedulable bool) error {
+	cordonPatch := []patchStringValue{{
```
kononovn, line 75: You use this var only once. Please use it directly as argument in Marshal func

kononovn (changes requested):
Please check my comments

PR #56, opened 2022-04-01
Updated fail test cases, added new uncertified operator

theemj:
-Updated some skip to cases to fail to reflect current tnf logic (test case change here: https://github.com/test-network-function/test-network-function/pull/661/files)
-added falcon operator to replace nginx,
which is no longer automatically deployed (https://github.com/test-network-function/cnf-certification-test-partner/pull/93/files)


sebrandon1 (approved):

kononovn (changes requested):
Please add logic which will handle the case when operators are already installed

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -93,52 +136,38 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
 
 			err = globalhelper.ValidateIfReportsAreValid(
 				affiliatedcertparameters.TestCaseOperatorAffiliatedCertName,
-				globalparameters.TestCaseSkipped)
+				globalparameters.TestCaseFailed)
 			Expect(err).ToNot(HaveOccurred(), "Error validating test reports")
 		})
 
 	// 46697
 	It("two operators to test, one belongs to certified-operators organization and its version is certified,"+
```
kononovn, line 144: Can you re-name this TC to something shorter? 
two operators to test, one is certified another is not certified [netative]

theemj: There are two different types of operators considered uncertified for this test:
- operators that are not in the certified-operators group in the Red Hat catalog
- operators that are in the certified-operators group but whose versions are not certified for the current ocp version

So there is a certain level of detail needed in the test case titles, but I can still try to make the ones like this shorter.

tests/affiliatedcertification/affiliatedcerthelper/affiliatedcerthelper.go
```diff
@@ -183,18 +185,32 @@ func updateCsv(namespace string, csv *v1alpha1.ClusterServiceVersion) error {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("Error deploying operator "+operatorPackage+": %w", err)
 	}
 
-	err = globalhelper.APIClient.ClusterServiceVersions(namespace).Delete(
-		context.TODO(), csv.GetName(), metav1.DeleteOptions{},
-	)
+	// confirm that operator is installed and ready
+	gomega.Eventually(func() bool {
```
kononovn, line 206: Please move Eventually call to tests package.  Let's avoid using gomega API calls in helpers.

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -54,30 +101,26 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
...
-				affiliatedcertparameters.ExistingOperatorNamespace,
+				affiliatedcertparameters.UncertifiedOperatorPrefixFalcon,
+				affiliatedcertparameters.TestCertificationNameSpace,
 				affiliatedcertparameters.OperatorLabel)
 			Expect(err).ToNot(HaveOccurred(), "Error labeling operator "+
-				affiliatedcertparameters.UncertifiedOperatorPrefixNginx)
+				affiliatedcertparameters.UncertifiedOperatorPrefixFalcon)
+
+			// add falcon operator info to array for cleanup in AfterEach
+			installedLabeledOperators = append(installedLabeledOperators, affiliatedcertparameters.OperatorLabelInfo{
```
kononovn, line 119: Is it possible to avoid using installedLabeledOperators between test cases and AfterEach section. It would be perfect if we could make test cases more isolated.

kononovn (changes requested):
Looks good, just few more comments to fix.

tests/affiliatedcertification/tests/affiliated_certification_operator.go
kononovn: Ok I see. In this case I would use something like:
two operators to test, the first is certified, the second is certified but has invalid version [netative]

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -35,13 +34,100 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
 		Expect(err).ToNot(HaveOccurred(), "Error creating namespace")
 
 		By("Deploy OperatorGroup")
```
kononovn, line 36: Please re-write this By. You deploy operator only in case when the "if" condition below is true

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -35,13 +34,100 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
...
-		err = affiliatedcerthelper.DeployOperatorGroup(affiliatedcertparameters.TestCertificationNameSpace,
-			utils.DefineOperatorGroup("affiliatedcert-test-operator-group",
+		if affiliatedcerthelper.IsOperatorGroupInstalled(affiliatedcertparameters.OperatorGroupName,
```
kononovn, line 37: In this "If" condition IsOperatorGroupInstalled func is used however in the next couple of the same conditions another function is consumed IsOperatorInstalled.

kononovn (changes requested):
Please fix couple of comments above.

tests/affiliatedcertification/tests/affiliated_certification_operator.go
theemj: This section deals with the operator group object, while the next deals with the operators themselves.

kononovn (approved):

PR #72, opened 2022-04-28
Adaptations for new tnf repo

shaior:
This PR contains adaptations in all of the suites, to work with the LaunchTests func after 
deleting the skipRegex. 

also, it includes one more fix in the lifecycle - using globalhelper.Configuration.General.TestImage.


config/config.yaml
```diff
@@ -9,4 +9,4 @@ general:
   tnf_config_dir: "/tmp/tnf_config"
   tnf_report_dir: "/tmp/tnf_report"
   tnf_image: "quay.io/testnetworkfunction/cnf-certification-test"
-  tnf_image_tag: "latest"
+  tnf_image_tag: "unstable"
```
kononovn, line 12: Please move it back to the latest.

PR #85, opened 2022-05-21
Refactor and restructure operator cert tests

theemj:


tests/globalhelper/runhelper.go
```diff
@@ -64,6 +64,9 @@ func LaunchTests(testCaseName string, tcNameForReport string) error {
 		cmd.Stderr = outfile
 	}
 
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
```
shaior, line 68: @theemj These two lines are already happening inside the logging files, we should stick to the logging files since they are not subject to change like our stdout.

tests/affiliatedcertification/affiliatedcerthelper/operator.go
```diff
@@ -201,3 +201,18 @@ func IsCatalogSourceEnabled(name, namespace, displayName string) bool {
...
+func DeleteCatalogSource(name, namespace, displayName string) error {
+	source, err := globalhelper.APIClient.CatalogSources(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if source.Spec.DisplayName == displayName {
+		err = globalhelper.APIClient.Delete(context.TODO(), source)
+
+		return err
```
shaior, line 214: @theemj Since the err would be nil if no error has occurred, I don't see any reason to return twice, I would just do the if, and then return err. WDYT?

tests/affiliatedcertification/tests/affiliated_certification_invalid_operator.go
```diff
@@ -138,22 +140,24 @@ var _ = Describe("Affiliated-certification invalid operator certification,", fun
...
 
-			var installPlan *v1alpha1.InstallPlan
 			Eventually(func() bool {
-				installPlan, err = affiliatedcerthelper.GetInstallPlanByCSV(affiliatedcertparameters.TestCertificationNameSpace,
+				installPlan, err := affiliatedcerthelper.GetInstallPlanByCSV(affiliatedcertparameters.TestCertificationNameSpace,
 					affiliatedcertparameters.UncertifiedOperatorFullSriov)
 				if err == nil {
-					return installPlan.Status.Phase != "" &&
-						installPlan.Status.Phase != v1alpha1.InstallPlanPhasePlanning
+					if installPlan.Status.Phase != "" &&
```
shaior: @theemj perhaps we can drop the return true/ false, and just return installpan.status.phase != " "... since this is already a boolean, WDYT?

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -97,22 +97,24 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
...
 				if err == nil {
-					return installPlan.Status.Phase != "" &&
-						installPlan.Status.Phase != v1alpha1.InstallPlanPhasePlanning
+					if installPlan.Status.Phase != "" &&
+						installPlan.Status.Phase != v1alpha1.InstallPlanPhasePlanning &&
+						installPlan.Status.Phase != v1alpha1.InstallPlanPhaseInstalling {
+						_ = affiliatedcerthelper.ApproveInstallPlan(affiliatedcertparameters.TestCertificationNameSpace,
+							installPlan)
+
+						return true
```
shaior: @theemj same here.

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -146,23 +148,25 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
...
```
shaior: @theemj same here.

tests/affiliatedcertification/affiliatedcerthelper/operator.go
```diff
@@ -205,7 +205,7 @@ func IsCatalogSourceEnabled(name, namespace, displayName string) bool {
 func DeleteCatalogSource(name, namespace, displayName string) error {
 	source, err := globalhelper.APIClient.CatalogSources(namespace).Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil {
-		return err
+		return nil
```
shaior, line 208: @theemj  I saw your comment on this commit, but something with this line is not clear to me, maybe you meant to return the err?

tests/affiliatedcertification/affiliatedcerthelper/operator.go
```diff
@@ -205,7 +205,7 @@ func IsCatalogSourceEnabled(name, namespace, displayName string) bool {
...
 	}
 
 	if source.Spec.DisplayName == displayName {
```
shaior, line 211: @theemj  I think we can drop the 2 returns, and just return err, since it will be nil if no error has occurred,
WDYT?

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -343,3 +293,49 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
 	})
 
 })
+
+func waitUnitlOperatorIsReady(csvPrefix, namespace string) error {
```
shaior, line 297: waitUnitlOperatorIsReady -> waitUntilOperatorIsReady

tests/affiliatedcertification/tests/affiliated_certification_operator.go
```diff
@@ -343,3 +293,49 @@ var _ = Describe("Affiliated-certification operator certification,", func() {
...
+		}
+
+		return false
+	}, affiliatedcertparameters.Timeout, affiliatedcertparameters.PollingInterval).Should(Equal(true),
+		csvPrefix+" is not ready.")
+
+	return err
+}
+
+func arrpoveInstallPlanWhenReady(csvName, namespace string) error {
```
shaior, line 318: arrpoveInstallPlanWhenReady -> approveInstallPlanWhenReady

shaior (changes requested):
@theemj I left some comments, please take a look :)

tests/affiliatedcertification/affiliatedcerthelper/operator.go
shaior: I understood that if no catalogsource was found, then the err wouldn't be nil, thus I think it's better to return bool,error, incase no catalogsource was found, then return false,nil.

PR #94, opened 2022-06-02
Update dep gopkg

shaior:
Switched "gopkg.in/yaml.v2"  to "gopkg.in/yaml.v3" in go.mod, then ran go mod tidy. 
this is supposed to fix security alert in the QE repo.

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -221,7 +221,7 @@ gopkg.in/inf.v0
 # gopkg.in/yaml.v2 v2.4.0
 ## explicit; go 1.15
 gopkg.in/yaml.v2
-# gopkg.in/yaml.v3 v3.0.1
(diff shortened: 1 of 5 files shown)


vendor/modules.txt
```diff
@@ -216,6 +216,8 @@ google.golang.org/protobuf/types/known/timestamppb
 # gopkg.in/inf.v0 v0.9.1
 ## explicit
 gopkg.in/inf.v0
+# gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7
```
kononovn, line 219: It looks like you added only 2 commented stings in to the file. Are you sure it would fix the root issue ?

shaior: It affected more files, I hope it will help.
If it won't, we will revert this one.

kononovn: But you need to updated dependencies right ? Looks like the old yaml dep is still there

PR #99, opened 2022-06-08
Removed prefixes of local test suites' helpers/params.

greyerof:
All test suites use a globalhelper and local/ts ad hoc helper functions.
As it's a common pattern for this project, there's no need for using the
test suite name as prefix for folders, files or package names.

In order to differentiate them from the global ones, the prefix "ts" ("test
suite") can be reused across all the test suites.

In short:
- This commit avoids the redundant repetition of the ts name.
- Makes helper/params import names shorter, helping the code to not
  trespass on the 120 chars line width.


shaior (approved):
lgtm

kononovn (approved):
lgtm

PR #100, opened 2022-06-08
Automated TCs for the observability-crd-status test case.

greyerof:
- Added apiextensions interface to the go-client.
- Created globalhelpers for CRD creation/deletion.
- Added support for targetCrdFilters for the TNF yaml config file.


tests/globalhelper/crd.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		for _, condition := range runningCrd.Status.Conditions {
+			if condition.Type == apiextv1.Established {
+				return true
+			}
+		}
+
+		return false
+	}, timeout, crdRetryIntervalSecs*time.Second).Should(Equal(true), "CRD is not ready")
```
theemj, line 49: Might make sense to move the *time.Second up to the const definition since it's used twice in the same way

theemj (approved):
just one style comment  that wouldn't change the code functionality either way. looks good to me!

tests/globalhelper/crd.go
greyerof: Done.

tests/globalhelper/crd.go
```diff
@@ -0,0 +1,70 @@
+package globalhelper
```
kononovn, line 1: Function from this module is used only in observability suite. So better to keep them in observability/helper

kononovn (changes requested):
Looks good except 1 comment above.

tests/globalhelper/crd.go
greyerof: Done!

kononovn (approved):
lgtm

PR #114, opened 2022-07-20
Automated platform-alteration-hugepages-config tcs

shaior:
Automated 51308, 51309

### DIFF
--- a/tests/platformalteration/helper/helper.go
+++ b/tests/platformalteration/helper/helper.go
@@ -44,6 +44,7 @@ func WaitForSpecificNodeCondition(clients *client.ClientSet, timeout, interval t
 func UpdateAndVerifyHugePagesConfig(updatedHugePagesNumber int, filePath string, pod *corev1.Pod) (bool, error) {
 	cmd := fmt.Sprintf("echo %d > %s", updatedHugePagesNumber, filePath)
 	_, err := globalhelper.ExecCommand(*pod, []string{"/bin/bash", "-c", cmd})
+
 	if err != nil {
 		return false, err
 	}
@@ -67,13 +68,13 @@ func UpdateAndVerifyHugePagesConfig(updatedHugePagesNumber int, filePath string,
 
 		time.Sleep(tsparams.RetryInterval * time.Second)
 	}
-
 }
 
 // GetHugePagesConfigNumber returns hugepages config number from a given file.
 func GetHugePagesConfigNumber(file string, pod *corev1.Pod) (int, error) {
 	cmd := fmt.Sprintf("cat %s", file)
 	buf, err := globalhelper.ExecCommand(*pod, []string{"/bin/bash", "-c", cmd})
+
 	if err != nil {
 		return -1, err
 	}
--- a/tests/platformalteration/helper/helper.go
+++ b/tests/platformalteration/helper/helper.go
@@ -2,8 +2,14 @@ package helper
 
 import (
 	"context"
+	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
+	tsparams "github.com/test-network-function/cnfcert-tests-verification/tests/platformalteration/parameters"
+
+	"github.com/test-network-function/cnfcert-tests-verification/tests/globalhelper"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/client"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/nodes"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


theemj (approved):
lgtm!

tests/platformalteration/tests/platform_alteration_hugepages_config.go
```diff
@@ -0,0 +1,98 @@
...
+
+		podList, err := globalhelper.GetListOfPodsInNamespace(tsparams.PlatformAlterationNamespace)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Get first hugepages file")
+		nrHugepagesFiles, err := globalhelper.ExecCommand(
+			podList.Items[0], []string{"/bin/bash", "-c", tsparams.FindHugePagesFiles})
+		Expect(err).ToNot(HaveOccurred())
+
+		hugePagesPaths := strings.Split(nrHugepagesFiles.String(), "\r\n")
```
greyerof, line 69: It would be nice having a check for this slice not to be empty.

greyerof (approved):

tests/platformalteration/helper/helper.go
```diff
@@ -32,3 +38,51 @@ func WaitForSpecificNodeCondition(clients *client.ClientSet, timeout, interval t
 		return false, nil
 	})
 }
+
+// UpdateAndVerifyHugePagesConfig updates the hugepages file with a given number,
+// and verify that it was updated successfully.
+func UpdateAndVerifyHugePagesConfig(updatedHugePagesNumber int, filePath string, pod *corev1.Pod) (bool, error) {
```
greyerof, line 44: Is it really necessary to return both `bool` and `error` here? I believe you only need to return `error`, but it would require a custom error message depending on the case. To use this function, you could do like this:
```
err := tshelper.UpdateAndVerifyHugePagesConfig(updatedHugePagesNumber, hugePagesPaths[0], &podList.Items[0])
Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("failed to update and verify hugepages: %v", err))
```

tests/platformalteration/helper/helper.go
```diff
@@ -32,3 +38,51 @@ func WaitForSpecificNodeCondition(clients *client.ClientSet, timeout, interval t
...
+	cmd := fmt.Sprintf("echo %d > %s", updatedHugePagesNumber, filePath)
+	_, err := globalhelper.ExecCommand(*pod, []string{"/bin/bash", "-c", cmd})
+
```
greyerof, line 47: Just my recommendation for this kind of `err := functionCall()` and the following `if err != nil` checks:
```
	cmd := fmt.Sprintf("echo %d > %s", updatedHugePagesNumber, filePath)

	_, err := globalhelper.ExecCommand(*pod, []string{"/bin/bash", "-c", cmd})
	if err != nil {
		return false, err
	}

    timeout := time.Now().Add(5 * time.Minute)
```
It feels more natural (at least in go) to check the error in the following line right after the function call.
The empty line between the `cmd` variable assignment and the function call makes this block easier to read. That will also avoid the linting error you'll get otherwise...

tests/utils/crd/crd.go
```diff
@@ -59,3 +62,18 @@ func DefineCustomResourceDefinition(names apiextv1.CustomResourceDefinitionNames
 		},
 	}
 }
+
+func EnsureCrdExists(name string) (bool, error) {
+	apiextv1, err := globalhelper.APIClient.ApiextensionsV1Interface.CustomResourceDefinitions().Get(
+		context.Background(), name, metav1.GetOptions{})
+
+	if err != nil {
```
greyerof, line 70: Have you tested this? I think it won't work, as Get() will return an error if the CRD does not exist, and your TC won't be skipped. I think this and some other helper functions can be improved when they use the "Get()" function to get a single resource. You need to use the errors package from the k8s apimachinery:
```
k8sErrors "k8s.io/apimachinery/pkg/api/errors"

func EnsureXXXResourceExists(name string) (bool, error) {
  _, err := globalhelper.APIClient.XXX.Get(...)
  if err != nil {
      if k8sErrors.IsNotFound(err) {
          return false, nil
      } else {
          return false, err
      }
  }
  
  return true, nil 
}
```
If the resource you're asking for does not exist, the Get() function will return a IsNotFound error. It's a bit weird, but as per your function, that's not an error but an expected output when the resource does not exist, so it makes more sense to me to return `false, nil` in that case. If the error is different, then it might be because any other unexpected scenario (faulty connection, permissions, wrong kubeconfig file...). In that case, this unexpected error should be propagated to the caller with `return false, err`.

tests/platformalteration/helper/helper.go
shaior: thanks, agree, usually i try to stick to this way, missed it.

tests/utils/crd/crd.go
shaior: thanks! agreed and fixed.

tests/utils/crd/crd.go
```diff
@@ -59,3 +63,22 @@ func DefineCustomResourceDefinition(names apiextv1.CustomResourceDefinitionNames
 		},
 	}
 }
+
+func EnsureCrdExists(name string) (bool, error) {
+	crd, err := globalhelper.APIClient.ApiextensionsV1Interface.CustomResourceDefinitions().Get(
```
greyerof, line 68: Minor comment: you can omit the `ApiextensionsV1Interface` part:
```
crd, err := globalhelper.APIClient.CustomResourceDefinitions().Get(
...
```

greyerof (approved):

PR #115, opened 2022-07-21
Access control pod automount service account token

theemj:
infrastructure + four test cases

### DIFF
--- a/tests/accesscontrol/helper/helper.go
+++ b/tests/accesscontrol/helper/helper.go
@@ -1,11 +1,16 @@
 package helper
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
+	"github.com/test-network-function/cnfcert-tests-verification/tests/accesscontrol/parameters"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/globalhelper"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/client"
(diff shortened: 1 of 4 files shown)


tests/accesscontrol/helper/helper.go
```diff
@@ -28,3 +33,21 @@ func DeleteNamespaces(nsToDelete []string, clientSet *client.ClientSet, timeout
 
 	return nil
 }
+
+func DefineDeployment(replica int32, containers int, name string) (*v1.Deployment, error) {
```
shaior, line 37: I think we have the same function in lifecycle/helper, perhaps we can create this function on a global level 
with a more meaningful name, and use it from there?

shaior (commented):
LGTM, left one comment.

tests/accesscontrol/helper/helper.go
theemj: Since both the DefineDeployment functions in lifecycle/helper and access-control/helper call DefineDeployment from the utils/deployment package with test suite specific values, I think it is probably fine to leave them in their own helper packages. We could replace the helper functions with calls to the utils/deployment version, but then we'd have to repeat a lot of args that will be the same for each test.

shaior (approved):
lgtm

greyerof (approved):

PR #116, opened 2022-07-25
Update operator-framework to latest

sebrandon1:
https://github.com/operator-framework/operator-sdk/releases


rdavid (approved):

theemj:
While testing this change I found that it causes one of the operators we use to not install correctly, resulting in the failure of two affiliated-certification-operator-is-certified test cases. Not sure of the exact cause, but it is the operator we intentionally install an older version of so that could be related. Continuing to look into this to see if there's a fix.

greyerof (approved):

sebrandon1:
I just rebased this PR.  Is it still on hold because of potential older version of the operator?

sebrandon1:
Closing this.  Will revisit in another PR.

PR #119, opened 2022-08-01
removed and re-added label from operator used by other test suite

theemj:
Removed and re-added label from operator used by other test suite before and after affiliated-certification suite.

### DIFF
--- a/tests/affiliatedcertification/affiliated_certification_suite_test.go
+++ b/tests/affiliatedcertification/affiliated_certification_suite_test.go
@@ -5,7 +5,6 @@ package affiliatedcertification
 import (
 	"flag"
 
-	"github.com/golang/glog"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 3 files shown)


tests/affiliatedcertification/affiliated_certification_suite_test.go
```diff
@@ -50,4 +58,10 @@ var _ = AfterSuite(func() {
 	By("Remove reports from report directory")
 	err = globalhelper.RemoveContentsFromReportDir()
 	Expect(err).ToNot(HaveOccurred())
+
+	By("Re-label operator used in other suites")
```
theemj, line 62: the reason there's no Expect(err).ToNot(HaveOccurred()) on this or the un-labeling above is that the cloudcasa operator is not installed on every ocp version, so we don't want the suite to fail if it's not there to be labeled.

greyerof: I don't like too much this approach. I think you should only re-apply the label if the operator actually had it before your test suite.

theemj: Added a check :+1:

tests/affiliatedcertification/helper/helper.go
```diff
@@ -103,6 +103,21 @@ func DeleteLabelFromInstalledCSV(prefixCsvName string, namespace string, label m
...
 }
 
+func DoesOperatorHaveLabel(prefixCsvName string, namespace string, label map[string]string) bool {
+	csv, err := GetCsvByPrefix(prefixCsvName, namespace)
+	if err != nil {
+		return false
+	}
+
+	for k := range csv.GetLabels() {
+		if _, ok := label[k]; ok {
```
greyerof, line 113: There are a couple of issues with this check. First: labels have the format "key=value", but you're only checking the key here, because "k" has each of the keys of the map[string]string from csv.GetLabels(). Second, this function is called "DoesOperatorHaveLabel", but the label is a map, which can actually have more than one label. Besides, the loop returns when only one of the csv's keys exists in the `label` map. I would add a "s" to the name of the function and to the label parameter, and would invert the checking order:
```
func DoesOperatorHaveLabels(..., labels map[string]string) ... {
    ...
	csvLabels := csv.GetLabels()
	for k, v := range labels {
		csvLabelValue, csvLabelExists := csvLabels[k]
		if !csvLabelExists || csvLabelValue != v {
			return false
		}
	}

	return true
}
```
This should work for whatever number of labels the parameter has. What do you think about it?

theemj: good point, I think this makes sense

greyerof (approved):

PR #123, opened 2022-08-09
Access control pod host ipc tests

theemj:


tests/utils/deployment/deployment.go
```diff
@@ -235,6 +235,12 @@ func RedefineWithPriviledgedContainer(deployment *v1.Deployment) *v1.Deployment
 	return deployment
 }
 
+func RedefineWithHostIpc(deployment *v1.Deployment, hostIpc bool) *v1.Deployment {
```
sebrandon1, line 238: Since `deployment` is already a pointer/passed by reference, you don't have to return the pointer again.  Because what I think it's doing is making a copy.

https://go.dev/play/p/eLHe_Oy0VjH

In my go playground it comes out to be two separate memory addresses.

sebrandon1 (commented):
Just one comment about pointers, other than that I think it's good.

tests/utils/deployment/deployment.go
theemj: ah good catch. I updated RedefineWithHostIpc and RedefineWithAutomountServiceAccountToken because I was using that one in the exact same way :+1:

sebrandon1 (approved):
Looks good 👍

shaior (approved):

PR #126, opened 2022-08-17
Add 2 platform-alteration-boot-params tests

shaior:
- Added 2 boot params tcs 51302, 51305.

- The negative tc (51306) is not stable, I think it might not be suitable for automation, I will keep investigating for a solution.
the files /proc/cmdline, /boot/loader/entries are not writeable, thus had to try with an rpm-ostree command, but that seems to be not stable enough.

- Had to change golang linter (lll) and limit it to 135 chars a line, due to the length of machine config import in client/client.go.


tests/platformalteration/tests/platform_alteration_boot_params.go
```diff
@@ -0,0 +1,81 @@
...
+		for _, machineConfig := range machineConfigList.Items {
+			for _, mcp := range machineConfigPoolList.Items {
+				if machineConfig.Name == mcp.Spec.Configuration.Name && mcp.Name == "worker-cnf" {
+					machineConfig.Spec.KernelArguments = []string{"skew_tick=1", "nohz=off"}
+
+					By("Update the current machineConfig")
+					_, err := globalhelper.APIClient.MachineConfigs().Update(context.TODO(), &machineConfig, metav1.UpdateOptions{})
+					Expect(err).ToNot(HaveOccurred())
+
+					By("Start platform-alteration-boot-params test")
```
theemj, line 68: Do the test launch and validation steps need to be within these for loops?

theemj:
one question, otherwise lgtm :+1:

tests/platformalteration/tests/platform_alteration_boot_params.go
shaior: Right, good catch, thanks.

theemj (approved):

PR #131, opened 2022-09-05
Add 2 sysctl_config tests

shaior:
Automated  51302 +  51332.

The negative scenario (51333) is still under investigation, currently having issues creating a file in
/etc/sysctl.d/... that contains a key-value pair.


theemj (approved):
looks good to me!

sebrandon1 (approved):

PR #133, opened 2022-09-07
Changed deployment spec-changing funcs that don't use pointers for values to return deployment pointer

theemj:
Based on mostly process of elimination, I think this *might* fix the access-control tests that fail due to deployment creation timeout. If we merge it in and it doesn't fix those tests during the Jenkins build, I'll revert it.

### DIFF
--- a/tests/utils/deployment/deployment.go
+++ b/tests/utils/deployment/deployment.go
@@ -235,18 +235,24 @@ func RedefineWithPriviledgedContainer(deployment *v1.Deployment) *v1.Deployment 
 	return deployment
 }
 
-func RedefineWithHostPid(deployment *v1.Deployment, hostPid bool) {
+func RedefineWithHostPid(deployment *v1.Deployment, hostPid bool) *v1.Deployment {
 	deployment.Spec.Template.Spec.HostPID = hostPid
+
+	return deployment
 }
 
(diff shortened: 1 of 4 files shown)


tests/utils/deployment/deployment.go
```diff
@@ -235,18 +235,24 @@ func RedefineWithPriviledgedContainer(deployment *v1.Deployment) *v1.Deployment
 	return deployment
 }
 
-func RedefineWithHostPid(deployment *v1.Deployment, hostPid bool) {
+func RedefineWithHostPid(deployment *v1.Deployment, hostPid bool) *v1.Deployment {
```
sebrandon1, line 238: Looking at this... I still don't think we want to return the same pointer because it's already pass-by-reference.  🤔

theemj: Fair but it is the way several other "RedefineWith..." functions in this file work. And while that's not a \*great\* reason for these functions to do it too, the non-returning way isn't consistently working so I'm thinking there may be k8s intricacies in play that I don't fully comprehend (which I suppose means my justification for this change is that I don't fully know what's going on lol)

sebrandon1: Also fair 😄

shaior (approved):

sebrandon1 (approved):

PR #135, opened 2022-09-15
Add lifecycle pv relcaim tc, improve lifecycle suite readability

shaior:
Automated 54201, 54202, 54203, 54204, 54206, 54207

Add lifecycle pv relcaim tc, improve lifecycle suite readability and consistency.

### DIFF
--- a/tests/lifecycle/tests/lifecycle_persistent_volume_reclaim_policy.go
+++ b/tests/lifecycle/tests/lifecycle_persistent_volume_reclaim_policy.go
@@ -13,8 +13,6 @@ import (
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/namespaces"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/persistentvolume"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/persistentvolumeclaim"
-	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/pod"
-	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/replicaset"
 )
 
 var _ = Describe("lifecycle-persistent-volume-reclaim-policy", func() {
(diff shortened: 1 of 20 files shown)


tests/utils/replicaset/replicaset.go
```diff
@@ -38,3 +38,19 @@ func RedefineWithReplicaNumber(replicaSet *v1.ReplicaSet, replicasNumber int32)
 
 	return replicaSet
 }
+
+func RedefineWithVolume(replicaSet *v1.ReplicaSet, name string) *v1.ReplicaSet {
```
sebrandon1, line 42: I don't think you need to return the pointer here as it's already pass-by-reference.  Just modify the pointer in place.

tests/utils/pod/pod.go
```diff
@@ -57,3 +57,19 @@ func RedefineWithLivenessProbe(pod *corev1.Pod) *corev1.Pod {
 
 	return pod
 }
+
+func RedefineWithVolume(pod *corev1.Pod, name string) *corev1.Pod {
```
sebrandon1, line 61: Same here, I don't think you need to return the pointer.

sebrandon1 (commented):
Two comments.

tests/utils/pod/pod.go
shaior: correct, thanks.
Will fix and upload a new version, I will also implement this across other functions.

sebrandon1: Agreed, it can be addressed later just pointing it out.

shaior: I agree and i think it's better to do it in a different PR, 
I will start working on one.

theemj (approved):

tests/lifecycle/tests/lifecycle_persistent_volume_reclaim_policy.go
```diff
@@ -0,0 +1,253 @@
...
+		By("Define & create first pv")
+		persistentVolumea := persistentvolume.RedefineWithPVReclaimPolicy(
+			persistentvolume.DefinePersistentVolume(tsparams.TestPVName, tsparams.LifecycleNamespace), corev1.PersistentVolumeReclaimDelete)
+
+		err := tshelper.CreatePersistentVolume(persistentVolumea, tsparams.WaitingTime)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Define & create second pv")
+		persistentVolumeb := persistentvolume.RedefineWithPVReclaimPolicy(
+			persistentvolume.DefinePersistentVolume(tsparams.TestPVName, tsparams.LifecycleNamespace), corev1.PersistentVolumeReclaimRecycle)
```
greyerof, line 209: Is this really working? This PV has the same name as the first one (TestPVName).

shaior:
@greyerof as explained in the huddle, in the beforeEach block, I clean all resources that exist in the namespace, 
including PVs and PVCs. 
but, I was mistaken by relating the PVs deletion with the namespaces.clean, since PVs are cluster wide and not namespaced.
hence I moved the cleanPVs func to a different location.

As for the comment on the duplicated pv name, good catch, thank you, uploaded a new commit.

greyerof (approved):

PR #148, opened 2022-10-25
Automated affinity required pods tc verification

shaior:
Considering this [PR](https://github.com/test-network-function/cnf-certification-test/pull/608/files)
affinity-required-deployments/statefulSets will not be automated.

I had to edit the runhelper.go in order to be able to provide 1 out of 3 labels each tc must have: 
1. test case name
2. common/ extended tags
3. test suite name

thus, I added the -l argument in the cmd line, which will target the specific tc, instead of specifying if it's common/extended.

Eventually, using -l made the -f redundant, and hence I deleted it.

### DIFF
--- a/tests/lifecycle/parameters/parameters.go
+++ b/tests/lifecycle/parameters/parameters.go
@@ -49,7 +49,7 @@ var (
 		testPodLabelPrefixName: testPodLabelValue,
 		"app":                  "test",
 	}
-	AffinityRequired = map[string]string{
+	AffinityRequiredPodLabels = map[string]string{
 		"AffinityRequired": "true",
 	}
 )
--- a/tests/lifecycle/parameters/parameters.go
+++ b/tests/lifecycle/parameters/parameters.go
@@ -36,6 +36,7 @@ var (
 	TnfPersistentVolumeReclaimPolicyTcName = "lifecycle-persistent-volume-reclaim-policy"
 	TnfCPUIsolationTcName                  = "lifecycle-cpu-isolation"
 	TnfStartUpProbeTcName                  = "lifecycle-startup-probe"
+	TnfAffinityRequiredPodsTcName          = "lifecycle-affinity-required-pods"
 )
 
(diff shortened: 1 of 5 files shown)


tests/utils/pod/pod.go
```diff
@@ -25,7 +25,16 @@ func DefinePod(podName string, namespace string, image string) *corev1.Pod {
 
 // RedefinePodWithLabel adds label to given pod manifest.
 func RedefinePodWithLabel(pod *corev1.Pod, label map[string]string) {
-	pod.ObjectMeta.Labels = label
+	newMap := make(map[string]string)
```
greyerof, line 28: A couple of comments about your changes in this function:
1. Be aware that you're changing the implementation in a non-backwards-compatible way: This function was replacing all the pod labels with the function's parameter ones. Now, you're adding parameter's ones to the pod's existing ones. This might affect other test cases. Even if no other TC is affected by this change, I recommend, at least, to add this change to the PR description. Another option is just creating a new function for appending labels to pod's ones, like `pod.AppendLabels(pod *corev1.Pod, labels map[string]string)`.

2. Why do you need to create a new map to add the labels to the existing pod's map? I believe this code does the same:
```
func RedefinePodWithLabel(pod *corev1.Pod, labels map[string]string) {
  for name, value := range labels {
    pod.Labels[name] = value
  }
}
```

tests/lifecycle/parameters/parameters.go
```diff
@@ -48,4 +49,7 @@ var (
 		testPodLabelPrefixName: testPodLabelValue,
 		"app":                  "test",
 	}
+	AffinityRequired = map[string]string{
```
greyerof, line 52: Minor suggestion: I'll add the suffix "PodLabels" to this var name: `AffinityRequiredPodLabels`.

greyerof (approved):

sebrandon1 (approved):

PR #173, opened 2022-11-14
Added claim to reports folder

shaior:
In order to debug CI failures we need the claim reports as well.

### DIFF
--- a/tests/globalhelper/helper.go
+++ b/tests/globalhelper/helper.go
@@ -3,6 +3,7 @@ package globalhelper
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"strings"
@@ -276,4 +277,25 @@ func AllowAuthenticatedUsersRunPrivilegedContainers() error {
 	glog.V(5).Info("error to query RBAC policy")
 
 	return nil
+}
+
+// CopyFiles copy file from source to destination.
+func CopyFiles(src string, dst string) error {
+	originalFile, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+
+	defer originalFile.Close()
+
+	newFile, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+
+	defer newFile.Close()
+
+	_, err = io.Copy(newFile, originalFile)
+
+	return err
 }
--- a/tests/globalhelper/helper.go
+++ b/tests/globalhelper/helper.go
@@ -283,19 +283,22 @@ func AllowAuthenticatedUsersRunPrivilegedContainers() error {
 func CopyFiles(src string, dst string) error {
 	originalFile, err := os.Open(src)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to open file %s: %w", src, err)
 	}
 
 	defer originalFile.Close()
 
 	newFile, err := os.Create(dst)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create file %s: %w", dst, err)
 	}
 
 	defer newFile.Close()
 
 	_, err = io.Copy(newFile, originalFile)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


tests/globalhelper/helper.go
```diff
@@ -277,3 +278,24 @@ func AllowAuthenticatedUsersRunPrivilegedContainers() error {
 
 	return nil
 }
+
+// CopyFiles copy file from source to destination.
+func CopyFiles(src string, dst string) error {
+	originalFile, err := os.Open(src)
+	if err != nil {
+		return err
```
greyerof, line 286: Please, add more context to this error: `return fmt.Errorf("failed to open file %s: %w", src, err)`

tests/globalhelper/helper.go
```diff
@@ -277,3 +278,24 @@ func AllowAuthenticatedUsersRunPrivilegedContainers() error {
...
+	}
+
+	defer originalFile.Close()
+
+	newFile, err := os.Create(dst)
+	if err != nil {
+		return err
```
greyerof, line 293: Same here: `fmt.Errorf("failed to create file %s: %w", dst, err)`

tests/globalhelper/helper.go
```diff
@@ -277,3 +278,24 @@ func AllowAuthenticatedUsersRunPrivilegedContainers() error {
...
+	}
+
+	defer newFile.Close()
+
+	_, err = io.Copy(newFile, originalFile)
+
+	return err
```
greyerof, line 300: Same here.

tests/globalhelper/reporthelper.go
```diff
@@ -232,3 +232,19 @@ func containsString(list []string, item string) bool {
 
 	return false
 }
+
+func CopyClaimFileToTcFolder(tcName, formattedTcName string) {
+	srcClaim := path.Join(Configuration.General.TnfReportDir, globalparameters.DefaultClaimFileName)
+	dstClaim := path.Join(Configuration.General.ReportDirAbsPath, "Debug", getTestSuiteName(tcName), formattedTcName,
+		globalparameters.DefaultClaimFileName)
+
+	if _, err := os.Stat(srcClaim); err == nil {
```
greyerof, line 241: Minor suggestion to avoid nested "if"s:
```
_, err := err := os.Stat(srcClaim)
if err != nil {
    a) log(...)
    b) return fmt.Errorf() -> only if function is updated to return errors.
}

err = CopyFiles(srcClaim, dstClaim)
if err != nil {
   a) log(...)
   b) return fmt.Errorf(...) -> only if function is updated to return errors.
}

return (nil)
```

tests/globalhelper/reporthelper.go
```diff
@@ -232,3 +232,19 @@ func containsString(list []string, item string) bool {
 
 	return false
 }
+
+func CopyClaimFileToTcFolder(tcName, formattedTcName string) {
```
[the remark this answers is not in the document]
shaior: I agree, but this will cause the LaunchTests func to become more unclear about errors, 
in the future, refactoring should be done to this function fixing readability, and then we will apply this.

tests/globalhelper/reporthelper.go
shaior: thanks

tests/globalhelper/helper.go
shaior: thanks

tests/globalhelper/helper.go
shaior: thanks

tests/globalhelper/helper.go
shaior: thanks

greyerof (approved):

theemj (approved):

PR #179, opened 2022-11-17
Add ginkgo seed usage to improve debugging

shaior:
ginkgo seed is used to rerun the same exact job with a seed that was generated by ginkgo, which will improve debugging.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -3,7 +3,12 @@
 GOPATH="${GOPATH:-~/go}"
 export PATH=$PATH:$GOPATH/bin
 EXCLUDED_FOLDERS=""
+GINKGO_SEED_FLAG=""
 ALL_TESTS_FOLDERS=$(ls -d ./tests/*/)
+
+if [[ "${GINKGO_SEED_NUMBER}" != "" ]]; then 
+    GINKGO_SEED_FLAG="--seed=${GINKGO_SEED_NUMBER}"
+fi
 
 function run_tests {
     case $1 in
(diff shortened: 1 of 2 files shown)


scripts/run-tests.sh
```diff
@@ -3,8 +3,13 @@
 GOPATH="${GOPATH:-~/go}"
 export PATH=$PATH:$GOPATH/bin
 EXCLUDED_FOLDERS=""
+GINKGO_SEED_FLAG=""
 ALL_TESTS_FOLDERS=$(ls -d ./tests/*/)
 
+if [[ "${GINKGO_SEED_NUMBER}" != "" ]]; then 
+    GINKGO_SEED_FLAG="--seed=${GINKGO_SEED_NUMBER}"
```
greyerof, line 10: The [README.md](https://github.com/test-network-function/cnfcert-tests-verification/blob/main/README.md) doc should be updated to warn users about this new `GINKGO_SEED_NUMBER` env var. Maybe the ["How to debug"](https://github.com/test-network-function/cnfcert-tests-verification#how-to-debug) section is a good place for that info.

Also, it would be nice to have a message about the ginkgo's seed flag being used. Something like:
```
if [[ "${GINKGO_SEED_NUMBER}" != "" ]]; then
    echo "Using ginkgo seed number: ${GINKGO_SEED_NUMBER}"
    GINKGO_SEED_FLAG="--seed=${GINKGO_SEED_NUMBER}"
    ...
```

shaior: good one, thank you,

PR #185, opened 2022-12-07
Fixed security context in daemonSet.go

shaior:
1. Fixed security context in daemonSet.go
2. Fixed a tc that tried to deploy a NAD that uses an interface that exists only in worker nodes by disabling 
master scheduling, in that way, only worker nodes will be scheduled.

### DIFF
--- a/tests/networking/helper/helper.go
+++ b/tests/networking/helper/helper.go
@@ -137,11 +137,13 @@ func GetClusterMultusInterfaces() ([]string, error) {
 	var nodesInterfacesList [][]string
 
 	for _, runningPod := range podsList.Items {
-		nodeInterfaces, err := getInterfacesList(runningPod)
-		if err != nil {
-			return nil, err
+		if !strings.Contains(runningPod.Spec.NodeName, "master") {
+			nodeInterfaces, err := getInterfacesList(runningPod)
+			if err != nil {
+				return nil, err
+			}
+			nodesInterfacesList = append(nodesInterfacesList, nodeInterfaces)
 		}
-		nodesInterfacesList = append(nodesInterfacesList, nodeInterfaces)
 	}
 
 	var lastMatch []string
--- a/tests/networking/tests/networking_multus_links.go
+++ b/tests/networking/tests/networking_multus_links.go
@@ -1,6 +1,7 @@
 package tests
 
 import (
+	"fmt"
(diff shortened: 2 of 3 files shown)


tests/networking/helper/helper.go
```diff
@@ -137,11 +137,13 @@ func GetClusterMultusInterfaces() ([]string, error) {
 	var nodesInterfacesList [][]string
 
 	for _, runningPod := range podsList.Items {
-		nodeInterfaces, err := getInterfacesList(runningPod)
-		if err != nil {
-			return nil, err
+		if !strings.Contains(runningPod.Spec.NodeName, "master") {
```
greyerof, line 140: This looks like a weak way to check whether a node is master. I recommend checking for the kubernetes standard node role label (node-role.kubernetes.io/master) instead. This can be done in a further PR, though.

greyerof (approved):

tests/networking/helper/helper.go
shaior: Agreed, thanks

greyerof (approved):

tests/utils/nodes/nodes.go
```diff
@@ -93,3 +93,16 @@ func setUnSchedulableValue(clients *client.ClientSet, nodeName string, unSchedul
...
 	return nil
 }
+
+func IsNodeMaster(name string, clients *client.ClientSet) (bool, error) {
+	node, err := clients.Nodes().Get(context.Background(), name, metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	if _, exists := node.Labels["node-role.kubernetes.io/master"]; exists {
```
sebrandon1, line 103: I think you might have to handle `control-plane` as well. 

Similar to: https://github.com/test-network-function/cnf-certification-test/pull/512

sebrandon1: Also: https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/provider/provider.go#L62 for reference.

greyerof: Good catch!

shaior: Thanks ! fixed

sebrandon1 (approved):

PR #216, opened 2023-01-31
Added -c arg to func that calls tnf test script

theemj:
This is a bit of a workaround since the value doesn't actually point to a docker config file, but since our test environment doesn't have a docker config to point to and a value appears to be necessary to run the script, this is what I went with (and it does work, so that's something lol).


sebrandon1:
@theemj The `-c` arg is optional right now but will be enforced when https://github.com/test-network-function/cnf-certification-test/pull/631 merges.

You can "make up" a docker config something like:
```
 mkdir -p /home/runner/.docker
 touch ${PFLT_DOCKERCONFIG}
 echo '{ "auths": {} }' >> ${PFLT_DOCKERCONFIG}
```

I'm doing this in the github runner because github doesn't have a docker config either for the CI.

sebrandon1 (approved):

PR #238, opened 2023-02-28
Removed quotes around suite name variables

theemj:
Turns out this is what was preventing the tests from running.


sebrandon1 (approved):

sebrandon1:
```
./scripts/run-tests.sh:31:90: note: Double quote to prevent globbing and word splitting. [SC2086]
./scripts/run-tests.sh:47:87: note: Double quote to prevent globbing and word splitting. [SC2086]
```
I think you have to add a shellcheck ignore for those lines.

PR #256, opened 2023-03-14
Add replace directive for MCO and openshift/client-go

sebrandon1:
https://github.com/test-network-function/cnfcert-tests-verification/network/updates/626158061

This is still failing with: 

```
updater | INFO <job_626158061> Updating github.com/openshift/client-go from 0.0.0-20230120202327-72f107311084 to 3.9.0+incompatible
```

3.9.0+incompatible isn't a thing.  [Here](https://github.com/openshift/api/issues/1421) is another issue brought up by another project.


sebrandon1:
Apparently I was correct with the replace in #254.

PR #262, opened 2023-03-17
Fixed negative test cases for platform-lateration-is-redhat-release.

greyerof:
The negative TCs for this TNF test case failed because both deployment and statefulset couldn't be created and the QE tc timed-out.

It happened because the selected non-UBI image for this TC doesn't have a /bin/bash binary. The shell is linked to busybox in /bin/sh instead.

### DIFF
--- a/tests/platformalteration/helper/helper.go
+++ b/tests/platformalteration/helper/helper.go
@@ -11,7 +11,11 @@ import (
 
 	"github.com/test-network-function/cnfcert-tests-verification/tests/globalhelper"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/client"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/deployment"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/nodes"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/statefulset"
+
+	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


tests/platformalteration/helper/helper.go
```diff
@@ -114,3 +118,37 @@ func AppendIstioContainerToPod(pod *corev1.Pod, image string) {
...
+func DefineDeploymentWithNonUBIContainer() *v1.Deployment {
+	dep := deployment.DefineDeployment(tsparams.TestDeploymentName, tsparams.PlatformAlterationNamespace,
+		tsparams.NotRedHatRelease, tsparams.TnfTargetPodLabels)
+
+	// Workaround as this non-ubi test image needs /bin/sh (busybox) instead of /bin/bash.
+	deployment.RedefineWithContainerSpecs(dep, []corev1.Container{
+		{
+			Name:    "test",
+			Image:   tsparams.NotRedHatRelease,
+			Command: []string{"/bin/sh", "-c", "sleep INF"},
```
rdavid, line 132: `sleep` expects an amount of seconds? Probably `while :; do sleep 2073600; done` from here:
 https://stackoverflow.com/questions/2935183/bash-infinite-sleep-infinite-blocking

tests/platformalteration/helper/helper.go
```diff
@@ -11,7 +11,11 @@ import (
 
 	"github.com/test-network-function/cnfcert-tests-verification/tests/globalhelper"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/client"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/deployment"
 	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/nodes"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/utils/statefulset"
+
+	v1 "k8s.io/api/apps/v1"
```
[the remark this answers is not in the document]
greyerof: Done.

tests/platformalteration/helper/helper.go
rdavid: It is not a POSIX/UNIX/macOS: "Although portable POSIX scripts must give sleep a single non-negative integer argument without a suffix, GNU sleep also accepts two or more arguments, unit suffixes, and floating-point numbers in either the current or the C locale". Busybox based Alpine container doesn't support it:
 https://stackoverflow.com/questions/39729241/why-does-the-container-created-with-docker-run-d-alpine-sleep-infinity-goes

PR #269, opened 2023-03-20
Added istio operator installation

theemj:
Added istio operator installation before service mesh tests


rdavid (approved):
LGTM

sebrandon1:
Is there also a way to remove the service mesh and test with/without?  I feel like we should probably have both covered.

theemj:
> 

we could potentially break the negative tests out into another file with a BeforeAll that checks if istio is installed and uninstalls it if so

PR #277, opened 2023-03-23
Replaced infinibox operator with prometurbo

theemj:
replaced infinibox operator with prometurbo, which should pass on all ocp versions


sebrandon1 (approved):
Approved.  In the future maybe we need some logic to tie certain operators to certain OCP versions so we don't have to find just one that works for all?

theemj:
> Approved. In the future maybe we need some logic to tie certain operators to certain OCP versions so we don't have to find just one that works for all?

Yup definitely worth considering. We've also run into the issue of operator versions that were once in the certified catalog being replaced over time so there's a fair amount of mulling to be done about how to approach these test cases moving forward.

PR #279, opened 2023-03-27
Improve go installation script

bnshr:
This is an improvement of the go installation where you fix a Go version and check if the QE machine has higher go version, if lower then we install, other skip.
In addition, Darwin platform is supported in this script.


theemj (approved):
looks good and works great!

scripts/install-latest-go.sh
```diff
@@ -1,10 +1,20 @@
...
+    echo "Version is up to date"
+else
+    unameOut="$(uname -s)"
+    case "${unameOut}" in
+        Linux*)
+                    wget https://go.dev/dl/"${REQUIRED_GO_VERSION}".linux-amd64.tar.gz
+                    rm -rf /usr/local/go && tar -C /usr/local -xzf "${REQUIRED_GO_VERSION}".linux-amd64.tar.gz
+                    rm "${REQUIRED_GO_VERSION}".linux-amd64.tar.gz
+                    ;;
+        *)          echo "Install go with version ${REQUIRED_GO_VERSION}"
```
sebrandon1, line 18:
```suggestion
        *)          echo "Installed go with version ${REQUIRED_GO_VERSION}"
```

sebrandon1 (approved):
One comment about tense.  Add a new line at the end of the file.  Looks good.

scripts/install-latest-go.sh
bnshr: That is an imperative sentence. It is up to the user on non-linux platform to install go based on his platform.

shaior (approved):
LGTM

PR #287, opened 2023-04-04
Performance suite - Test Exclusive CPU pool

bnshr:
Includes : 

- A new test suite called `performance` 
- Test case : Exclusive CPU pool

### DIFF
--- a/tests/globalhelper/runhelper.go
+++ b/tests/globalhelper/runhelper.go
@@ -92,5 +92,9 @@ func getTestSuiteName(testCaseName string) string {
 		return globalparameters.AccessControlSuiteName
 	}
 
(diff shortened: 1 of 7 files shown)


sebrandon1 (approved):
Looks good.  As a follow up when this merges, can you make a PR to change this to `true`?

https://github.com/test-network-function/cnf-certification-test/blob/main/cnf-certification-test/identifiers/identifiers.go#L461

bnshr:
> Looks good. As a follow up when this merges, can you make a PR to change this to `true`?
> 
> https://github.com/test-network-function/cnf-certification-test/blob/main/cnf-certification-test/identifiers/identifiers.go#L461

Thanks - I was not aware of significance of that field. I would do that.

PR #305, opened 2023-04-26
Isolated cpu pool

bnshr:


tests/performance/performance_suite_test.go
```diff
@@ -59,3 +66,81 @@ var _ = AfterSuite(func() {
 	err = globalhelper.RemoveContentsFromReportDir()
 	Expect(err).ToNot(HaveOccurred())
 })
+
+func ConfigurePrivilegedServiceAccount(namespace string) error {
```
theemj, line 70: For consistency with the other test suites it might make sense to move this and getPrivilegedServiceAccountObjects into performance/helper/helper.go, since the x_suite_test files usually just contain the TestX function and Before and After Suite vars

theemj (approved):
Successfully tested on the QE cluster - works great! One optional stylistic suggestion,  otherwise looks good to me :+1:

PR #307, opened 2023-04-27
QE Test:  Exclusive cpu pool scheduling

bnshr:


tests/performance/tests/rt_exclusive_cpu_pool_scheduling_policy.go
```diff
@@ -0,0 +1,119 @@
...
+			err = globalhelper.LaunchTests(
+				tsparams.TnfRtExclusiveCPUPoolSchedulingPolicy,
+				globalhelper.ConvertSpecNameToFileName(CurrentSpecReport().FullText()))
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Verify test case status in Junit and Claim reports")
+			err = globalhelper.ValidateIfReportsAreValid(
+				tsparams.TnfRtExclusiveCPUPoolSchedulingPolicy,
+				globalparameters.TestCasePassed)
+			Expect(err).To(HaveOccurred())
```
theemj, line 42: should be Expect(err).ToNot(HaveOccurred())

bnshr: Done!

tests/performance/tests/rt_exclusive_cpu_pool_scheduling_policy.go
```diff
@@ -0,0 +1,119 @@
...
+var _ = Describe("performance-exclusive-cpu-pool-rt-scheduling-policy",
+	func() {
+
+		BeforeEach(func() {
+			By("Clean namespace before each test")
+			err := namespaces.Clean(tsparams.PerformanceNamespace, globalhelper.APIClient)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		FIt("One pod running in exclusive cpu pool and shared cpu scheduling policy", func() {
```
theemj, line 23:
```suggestion
		It("One pod running in exclusive cpu pool and shared cpu scheduling policy", func() {
```

theemj:
Seeing inconsistent results when testing these. Sometimes the tests all execute as expected, other times the two positive cases fail with:
```
[FAILED] {"CompliantObjectsOut":null,"NonCompliantObjectsOut":[{"ObjectType":"Container","ObjectFields":{"ContainerName":"rt-app","Namespace":"performance-ns","PodName":"performance-pod","ReasonForNonCompliance":"Internal error, err=unable to run nsenter due to : cannot execute command: \" ps -e -o pidns,pid,args \"  on container: rt-app pod: performance-pod ns: performance-ns err:command terminated with exit code 1"}}]}
  In [It] at: /usr/tnf/tnf-src/pkg/testhelper/testhelper.go:176 
```

bnshr:
@theemj Could it be reviewed again? We have already an issue created for the inconsistency of rt tests in the performance suite for further investigation of this issue that are happening in existing test cases. The solution would be applied to this test case with that.

theemj (approved):

PR #340, opened 2023-06-01
non root user test in access control 

aabughosh:


config/testconfig.yaml
```diff
@@ -8,10 +8,10 @@ general:
   tnf_entry_point_script: "run-tnf-container.sh"
   # Example
   # tnf_config_dir: "/Users/bmandal/rhdev/github.com/cnfcert-tests-verification/tnf_config"
-  tnf_config_dir: "/path/to/cnfcert-tests-verification/tnf-config"
+  tnf_config_dir: "/Users/amal/tnf/cnfcert-tests-verification/tnf-config"
```
theemj, line 11: probably want to change this back to the stand-in value

theemj (approved):
one small comment on testconfig, otherwise looks good to me!

PR #377, opened 2023-07-13
Install ginkgo from go.mod

sebrandon1:
Related to: https://github.com/test-network-function/cnf-certification-test/pull/827

Deletes the script and adjusts the Makefile to match the `cnf-certification-test` repo.


Makefile
```diff
@@ -35,7 +35,7 @@ test-features: update-go install-ginkgo
 	FEATURES="$(FEATURES)" ./scripts/run-tests.sh features
 
 install-ginkgo:
-	scripts/install-ginkgo.sh
+	go install "$$(awk '/ginkgo/ {printf "%s/ginkgo@%s", $$1, $$2}' go.mod)"
```
rdavid, line 38: Maybe single `$`? `$$` is the process ID.

sebrandon1: It's the way we have it in the cnf-certification-test repo:

https://github.com/test-network-function/cnf-certification-test/blob/main/Makefile#L118

I just made it match.

rdavid: It is a Makefile insanity:
 https://www.gnu.org/software/make/manual/html_node/Secondary-Expansion.html

I need time to process it :-)

rdavid (approved):

PR #386, opened 2023-07-19
Remove init func

sebrandon1:
Trying to add unit tests to the QE repo to solve a problem and it is trying to run the `init()` function which is populating a list of clients which is not needed during unit test time.  

Bonus:
- Fixed the `make unit-tests` to work with this change.
- Modified the yamllint linter config based on the `cnf-certification-test` repo.
- Check out the `cnf-certification-test` repo because we need the TNF_REPO_PATH var set for the unit tests.

### DIFF
--- a/tests/globalhelper/init.go
+++ b/tests/globalhelper/init.go
@@ -18,14 +18,14 @@ func GetAPIClient() *testclient.ClientSet {
 		return apiclient
 	}
 
-	tempClient, err := config.DefineClients()
+	var err error
+	apiclient, err = config.DefineClients()
+
 	if err != nil {
 		glog.Fatal(fmt.Sprintf("can not load api client. Please check KUBECONFIG env var - %s", err))
 	}
 
-	apiclient = tempClient
-
-	return tempClient
+	return apiclient
 }
 
 func GetConfiguration() *config.Config {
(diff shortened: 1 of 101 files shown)


tests/globalhelper/init.go
```diff
@@ -2,35 +2,26 @@ package globalhelper
...
 	if err != nil {
 		glog.Fatal(fmt.Sprintf("can not load api client. Please check KUBECONFIG env var - %s", err))
 	}
 
-	Configuration, err = config.NewConfig()
+	return c
+}
 
+func NewConfiguration() *config.Config {
+	Configuration, err := config.NewConfig()
```
edcdavid, line 21: Don't we need to follow a singleton pattern here? looks like the configuration is re-created everytime

tests/globalhelper/init.go
```diff
@@ -2,35 +2,26 @@ package globalhelper
...
-		Configuration = &config.Config{}
-
-		return
-	}
-
-	var err error
-	APIClient, err = config.DefineClients()
-
+func NewAPIClient() *testclient.ClientSet {
+	c, err := config.DefineClients()
```
edcdavid, line 12: Same here, the client seems to be re-created

tests/globalhelper/init.go
```diff
@@ -2,35 +2,43 @@ package globalhelper
...
-		return
+func GetAPIClient() *testclient.ClientSet {
+	if apiclient != nil {
+		return apiclient
 	}
 
-	var err error
-	APIClient, err = config.DefineClients()
-
+	tempClient, err := config.DefineClients()
```
edcdavid, line 21: why having tempClient and Configuration variable? why not:
var err error
apiclient, err = config.DefineClients()

edcdavid (approved):

tests/globalhelper/init.go
sebrandon1: Good call, was a leftover.  I'll update it.

PR #393, opened 2023-07-24
Add EnsureAllNodesAreLabeled func to lifecycle suite

sebrandon1:
Add `EnsureAllNodesAreLabeled` func to the lifecycle suite of tests (for now) to ensure that `worker-cnf` is on all of the workers prior to them running.

@theemj mentioned that the QE CI automatically makes this label happen on Jenkins but as far as running in Kind/etc it doesn't happen automatically as of yet.  This just ensures that the label is available.


rdavid:
Why have almost 900 files changed?

sebrandon1:
@rdavid This repo vendors its dependencies, so any change to the go.mod results in a vendor change.

tests/lifecycle/tests/lifecycle_container_shutdown.go
```diff
@@ -1,27 +1,41 @@
...
 
 	tshelper "github.com/test-network-function/cnfcert-tests-verification/tests/lifecycle/helper"
 	tsparams "github.com/test-network-function/cnfcert-tests-verification/tests/lifecycle/parameters"
 )
 
 var _ = Describe("lifecycle-container-shutdown", func() {
 
+	configSuite, err := config.NewConfig()
+	if err != nil {
+		glog.Fatal(fmt.Errorf("can not load config file"))
```
rdavid, line 25: Should `err` be added to the message?

tests/lifecycle/tests/lifecycle_affinity_required_pods.go
```diff
@@ -31,6 +32,10 @@ var _ = Describe("lifecycle-affinity-required-pods", func() {
 		By("Clean namespace before each test")
 		err = namespaces.Clean(tsparams.LifecycleNamespace, globalhelper.GetAPIClient())
 		Expect(err).ToNot(HaveOccurred())
+
+		By("Ensure all nodes are labeled with 'worker-cnf' label")
```
edcdavid, line 36: optional(not sure that this would follow current code pattern): it seems that we are missing a "BeforeEach" and "AfterEach" at the program level before each test. Most of these are the same for every suite. For each suite, we could rename the namespace param to just "Namespace" from "(tsparams.LifecycleNamespace" to "tsparams.Namespace" . The import path makes this unique. Then implement functions globalHelper.BeforeEach(tsparams.Namespace) and globalHelper.AfterEach(tsparams.Namespace) with everything we need to do globally in every "BeforeEach" and "AfterEach". If some suite specific behavior, then add an "if" statement on the namespace inside these 2 new functions.

sebrandon1: This could probably be experimented with in another branch/PR.  However on the flip side I sort of like that each test is explicit with what it needs for pre-requisites for bring-up and tear-down.

edcdavid (approved):

tests/lifecycle/tests/lifecycle_container_shutdown.go
sebrandon1: Done.

rdavid (approved):

PR #394, opened 2023-07-24
Add 'make test' path to the Makefile

sebrandon1:
All this does is call the `unit-tests` but allows for the shorthand to run the unit tests.


Makefile
```diff
@@ -24,6 +24,8 @@ deps-update:
 	go mod tidy && \
 	go mod vendor
 
+test: unit-tests
+
```
rdavid, line 28: Why are the identical paths needed? It could be `test` only.

theemj (approved):

Makefile
sebrandon1: Didn't want to break any existing tooling that uses `make unit-tests` somewhere.

rdavid (approved):

PR #395, opened 2023-07-24
Modify EnableMasterScheduling

sebrandon1:
Instead of using `OcpClientInterface` to modify schedulers (OCP only), switch the `EnableMasterScheduling` function to use the k8s client and modify the node objects to remove the taint that is preventing scheduling in the first place.

Getting rid of the `OcpClientInterface` makes the QE code more compatible with running against plain k8s.

### DIFF
--- a/tests/globalhelper/nodes.go
+++ b/tests/globalhelper/nodes.go
@@ -8,10 +8,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	controlPlaneTaintKey = "node-role.kubernetes.io/control-plane"
+	masterTaintKey       = "node-role.kubernetes.io/master"
+)
+
 // EnableMasterScheduling enables/disables master nodes scheduling.
 func EnableMasterScheduling(scheduleable bool) error {
 	// Get all nodes in the cluster
-	nodes, err := GetAPIClient().CoreV1Interface.Nodes().List(context.Background(), metav1.ListOptions{})
+	nodes, err := GetAPIClient().CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
 		return fmt.Errorf("failed to get nodes: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


tests/globalhelper/nodes.go
```diff
@@ -4,22 +4,78 @@ import (
...
+}
+
+func addControlPlaneTaint(node *corev1.Node) error {
+	// add the control-plane:NoSchedule taint to the master
+	node.Spec.Taints = append(node.Spec.Taints, corev1.Taint{
+		Key:    "node-role.kubernetes.io/control-plane",
+		Effect: corev1.TaintEffectNoSchedule,
+	})
+
+	_, err := GetAPIClient().CoreV1Interface.Nodes().Update(context.Background(), node, metav1.UpdateOptions{})
```
edcdavid, line 46: is context.TODO() same as context.Background()? not sure.. Is the code waiting for this change to be applied somewhere else?

edcdavid (approved):

tests/globalhelper/nodes.go
sebrandon1: Whoops, yes they are the exact same function.  I'll switch this to TODO().

sebrandon1: Actually there are a ton of `context.Background()` references in the code and I can do that in another PR to keep everything separate.

tests/globalhelper/nodes.go
```diff
@@ -4,22 +4,78 @@ import (
...
 	return nil
 }
+
+func removeControlPlaneTaint(node *corev1.Node) error {
+	// remove the control-plane:NoSchedule taint from the master
+	for i, taint := range node.Spec.Taints {
+		if taint.Key == "node-role.kubernetes.io/master" || taint.Key == "node-role.kubernetes.io/control-plane" {
+			node.Spec.Taints = append(node.Spec.Taints[:i], node.Spec.Taints[i+1:]...)
+		}
+	}
```
rdavid, line 61: Append inside cycle by the slice changes the slice, right? Probably, it works here, but the idea doesn't feel right. Maybe:
```suggestion
	var updatedTaints []corev1.Taint
	for _, taint := range node.Spec.Taints {
		if !(taint.Key == "node-role.kubernetes.io/master" || taint.Key == "node-role.kubernetes.io/control-plane") {
			updatedTaints = append(updatedTaints, taint)
		}
	}
	node.Spec.Taints = updatedTaints
```

Also I'd move the hard-coded keys to const global slice.

sebrandon1: Added `const` vars and switched to `Switch` statement.

theemj (approved):

PR #397, opened 2023-07-25
Check for existing runtime class to avoid error

sebrandon1:
Issuing the `Create()` without checking first if runtimeClass exists ends in an error if it has already been created.

image

### DIFF
--- a/tests/globalhelper/runtimeclass.go
+++ b/tests/globalhelper/runtimeclass.go
@@ -12,7 +12,19 @@ import (
 )
 
 func CreateRunTimeClass(rtc *nodev1.RuntimeClass) error {
-	rtc, err := GetAPIClient().RuntimeClasses().Create(context.TODO(), rtc, metav1.CreateOptions{})
+	// Check for existing runtimeclass first
+	rtcCreated, err := isRtcCreated(rtc)
+	if err != nil {
+		return fmt.Errorf("failed to check if runtimeclass %q (ns %s) exists: %w", rtc.Name, rtc.Namespace, err)
+	}
+
+	if rtcCreated {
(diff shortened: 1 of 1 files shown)


tests/globalhelper/runtimeclass.go
```diff
@@ -12,7 +12,19 @@ import (
...
-	rtc, err := GetAPIClient().RuntimeClasses().Create(context.Background(), rtc, metav1.CreateOptions{})
+	// Check for existing runtimeclass first
+	rtcCreated, err := isRtcCreated(rtc)
+	if err != nil {
+		return fmt.Errorf("failed to check if runtimeclass %q (ns %s) exists: %w", rtc.Name, rtc.Namespace, err)
+	}
+
+	if rtcCreated {
+		glog.V(5).Info(fmt.Sprintf("runtimeclass %q (ns %s) already exists", rtc.Name, rtc.Namespace))
+
```
rdavid, line 23: There is no need for the vertical space.

tests/globalhelper/runtimeclass.go
```diff
@@ -12,7 +12,19 @@ import (
 )
 
 func CreateRunTimeClass(rtc *nodev1.RuntimeClass) error {
-	rtc, err := GetAPIClient().RuntimeClasses().Create(context.Background(), rtc, metav1.CreateOptions{})
+	// Check for existing runtimeclass first
+	rtcCreated, err := isRtcCreated(rtc)
```
rdavid, line 16: Does `Eventually` do the same below?

sebrandon1: Yeah it's a re-use of the same function that the Eventually uses.

rdavid (approved):

PR #407, opened 2023-07-31
Use HOME as default for the configurations

rdavid:


tests/utils/config/config.go
```diff
@@ -183,6 +183,10 @@ func readFile(cfg *Config, cfgFile string) error {
 		return err
 	}
 
+	cfg.General.TnfConfigDir = os.ExpandEnv(cfg.General.TnfConfigDir)
+	cfg.General.TnfReportDir = os.ExpandEnv(cfg.General.TnfReportDir)
+	cfg.General.DockerConfigDir = os.ExpandEnv(cfg.General.DockerConfigDir)
```
sebrandon1, line 188: These aren't needed because we are already using `envconfig` for these vars.

rdavid: Doesn't work without it. What is envconfig?

PR #438, opened 2023-08-21
Exclude SELinux-related QE tests on a local kind cluster

rdavid:
Test access_control_security_context verifies the SCC features of a cluster. However, configuring this behavior on a local kind cluster might not be feasible. Therefore, it's advisable to skip these tests when running on a local cluster. [Jira ticket](https://issues.redhat.com/browse/CNFCERT-613).


sebrandon1:
I like it.  I have a feeling we'll be using this function in the `platform` suite a lot.

tests/accesscontrol/tests/access_control_security_context.go
```diff
@@ -87,6 +91,10 @@ var _ = Describe("Access-control security-context,", func() {
 
 	// 63738
 	It("two deployments, one pod each, one container each, both have allowed security context", func() {
+		if !tshelper.IsKindCluster() {
```
sebrandon1, line 94: I think the logic should remove the !

rdavid: :-) The logic and ginkgo are on opposing sides, "should do something, if it can":
 https://onsi.github.io/ginkgo/#skipping-specs

rdavid: I'm particularly concerned about executing the command multiple times. Is there a way to ensure it runs once without parallel execution? Alternatively, should we fully commit and utilize `sync.Once`?

sebrandon1: Not too worried about it.  We can switch to a singleton or something if performance is impacted later.

sebrandon1 (dismissed):

sebrandon1:
@rdavid Can we start a list of non applicable tests in the README or another markdown file that lists the tests that are not compatible with a Kind environment?  Just so we have it listed publicly.

theemj (approved):

tests/accesscontrol/tests/access_control_security_context.go
sebrandon1: Also I think you still have to remove the ! because if it's a Kind cluster, we want to skip.  Right now the way you have it makes me think if its not a kind cluster, skip.

sebrandon1 (changes requested):
Change logic to match because it seems like this would be skipped on all OCP runs (or at least that's how I read it).

tests/accesscontrol/tests/access_control_security_context.go
rdavid: I've tested it, `!` is needed.

rdavid:
>

rdavid:
> Can we start a list of non applicable tests
Done.

PR #444, opened 2023-08-23
Add 'Serial' delimiter to Describe blocks

sebrandon1:
Adding `Serial` to the `Describe` sections of ginkgo blocks for future parallel runs to be enabled.  Adding the `Serial` flag to these blocks prevents them from running in parallel if `-p` is enabled (in the future it will be).

https://onsi.github.io/ginkgo/#serial-specs

When we enable parallel runs, it will be as easy as remove the `Serial` from the `Describe` line once we get the individual test suites setup to run in parallel.

PR #445, opened 2023-08-24
Enable parallel networking suite

sebrandon1:
Enables parallel runs in the `networking` suite!  🎉   This brings the amount of time to run the networking suite down from 30 minutes to ~80 seconds.  🤯 

image

Notable changes for reviewers:
- Added a function in the `globalhelper` package named `GenerateRandomString` which helps randomize the namespaces in which the resources are deployed into.
- Added the `-p` to the calls to ginkgo in `scripts/run-tests.sh`.  This enables the parallel running.  Follow-up to #444.
- Added `OverrideReportDir` and `OverrideTnfConfigDir` funcs for overriding the default directories on the machine running the QE tests.
- Moved the `"Ensure all nodes are labeled with 'worker-cnf' label"` step to the `BeforeSuite` block for organization.

### DIFF
--- a/tests/networking/helper/helper.go
+++ b/tests/networking/helper/helper.go
@@ -26,16 +26,16 @@ import (
 )
 
 // DefineAndCreateDeploymentOnCluster defines deployment resource and creates it on cluster.
-func DefineAndCreateDeploymentOnCluster(replicaNumber int32) error {
-	deploymentUnderTest := defineDeploymentBasedOnArgs(tsparams.TestDeploymentAName, tsparams.TestNetworkingNameSpace,
+func DefineAndCreateDeploymentOnCluster(replicaNumber int32, namespace string) error {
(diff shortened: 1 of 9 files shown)


theemj (approved):

tests/globalhelper/helper.go
```diff
@@ -361,3 +363,19 @@ func CopyFiles(src string, dst string) error {
...
 	return nil
 }
+
+var seededRand *rand.Rand = rand.New(
+	rand.NewSource(time.Now().UnixNano()))
+
+func GenerateRandomString(length int) string {
+	charset := "abcdefghijklmnopqrstuvwxyz"
+
+	//nolint:varnamelen
```
edcdavid, line 373: same comments as in https://github.com/test-network-function/cnfcert-tests-verification/pull/446
why not pull all these utility function in a new package and add new ones for before and after each?

PR #456, opened 2023-08-25
Enable parallel lifecycle suite

sebrandon1:
image

Notable changes:
- Introduced the `disable_intrusive_tests` boolean in to the config to be able to turn off the intrusive tests when needed.  The "plan" is to disable the intrusive tests during the QE PR runs, then enable them during nightlies.  The default value is `false` so the tests are on by default as to not break existing QE pipelines.
- Moved the helper func `IsKindCluster` from the accesscontrol helper to the `globalhelper` package.
- Brings back the `EnableMasterScheduling` functions removed in #437 as it is needed for Kind environments.  I have wrappers around this function checking for `IsKindCluster` first.
- The intrusive scaling tests and the pod-recreation tests are still ran in `Serial` as well as the reclaim policy tests.  For some reason, the default storageclass that is supplied by k8s (or lack thereof) doesn't like running in parallel and we can enable it later.


tests/utils/nodes/nodes.go
```diff
@@ -144,3 +149,81 @@ func LabelNode(client corev1Typed.CoreV1Interface, node *corev1.Node, label, val
 
 	return err
 }
+
+// EnableMasterScheduling enables/disables master nodes scheduling.
+func EnableMasterScheduling(client corev1Typed.CoreV1Interface, scheduleable bool) error {
```
edcdavid, line 154: This function is no longer needed and was removed by https://github.com/test-network-function/cnfcert-tests-verification/pull/437

tests/lifecycle/tests/lifecycle_affinity_required_pods.go
```diff
@@ -11,43 +11,69 @@ import (
...
 
-		By("Clean namespace before each test")
-		err = namespaces.Clean(tsparams.LifecycleNamespace, globalhelper.GetAPIClient())
+		By(fmt.Sprintf("Create %s namespace", randomNamespace))
+		err := namespaces.Create(randomNamespace, globalhelper.GetAPIClient())
 		Expect(err).ToNot(HaveOccurred())
 
-		By("Ensure all nodes are labeled with 'worker-cnf' label")
-		err = nodes.EnsureAllNodesAreLabeled(globalhelper.GetAPIClient().CoreV1Interface, configSuite.General.CnfNodeLabel)
+		By("Override default report directory")
```
edcdavid, line 37: same comment as for https://github.com/test-network-function/cnfcert-tests-verification/pull/446. create  common functions for BeforeEach and AfterEach

tests/utils/nodes/nodes.go
sebrandon1: Still needed for Kind cluster setups.  The calls to these funcs are wrapped in the `IsKindCluster` func.

PR #458, opened 2023-08-29
Bug fix: disallow global parallel run of the Ginkgo suites

jmontesi:
Not all the suites are ready to be launched in parallel.


sebrandon1:
I will verify the `Serial` flags are doing what they are supposed to be doing from: https://github.com/test-network-function/cnfcert-tests-verification/pull/444

sebrandon1 (approved):

PR #469, opened 2023-08-31
observability: add two more cases to pod-disruption-budget

jmontesi:
The new scenarios covered are:

* The PDB matchLabels field does not match the workload label

* There is no PDB defined

### DIFF
--- a/tests/observability/parameters/parameters.go
+++ b/tests/observability/parameters/parameters.go
@@ -24,6 +24,9 @@ const (
 
 	TestPodLabelKey   = "tnf-qe/observability"
 	TestPodLabelValue = "container-logging-tc"
+
+	UnknownKey   = "unknown-key"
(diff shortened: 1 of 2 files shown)


theemj:
lgtm. does "pod disruption budget matchLabels does not match deployment label" depend on https://github.com/test-network-function/cnf-certification-test/pull/1374 ? If so I imagine we should wait for that to be merged before merging these.

jmontesi:
> lgtm. does "pod disruption budget matchLabels does not match deployment label" depend on [test-network-function/cnf-certification-test#1374](https://github.com/test-network-function/cnf-certification-test/pull/1374) ? If so I imagine we should wait for that to be merged before merging these.

Correct.

sebrandon1 (approved):

theemj (approved):

PR #470, opened 2023-08-31
Fix the cleanup

sebrandon1:
#465 didn't clear out the contents of the directory first and was left with errors saying how the directory wasn't empty.


sebrandon1:
image

PR #472, opened 2023-08-31
Use DeferCleanup() in networking suite

sebrandon1:
Use ginkgo's `DeferCleanup` to clean these namespaces up incase of failure.


sebrandon1:
I found this was a thing in #471

theemj (approved):

PR #475, opened 2023-09-01
Add minReadySeconds to specs

sebrandon1:
In the `globalhelper` package, we have a bunch of these funcs that `CreateAndWait...` and for the most part I think they do their job waiting for the number of Running replicas to be "up" but I don't think the resources are truly up and running before the tests are kicked off which is generally right after we create the resources.

Adding these sleeps seems to fix the instability I've been seeing in my test runs:
https://github.com/test-network-function/cnf-certification-test/actions/runs/6053059734?pr=1377 (for example)

EDIT:

I have taken @jmontesi's advice and switched to `minReadySeconds: 30` for deployments, daemonsets, and replicasets.  This seems to have cleared up the random failures I have been seeing in my testing.  I also changed the conditional statements for how to determine if a resource is "ready".  See the individual files for the changes.

### DIFF
--- a/tests/globalhelper/daemonset.go
+++ b/tests/globalhelper/daemonset.go
@@ -50,7 +50,7 @@ func isDaemonSetReady(namespace string, name string) (bool, error) {
 		return false, err
 	}
 
-	if daemonSet.Status.NumberReady > 0 && daemonSet.Status.NumberUnavailable == 0 {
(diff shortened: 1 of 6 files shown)


theemj (approved):

rdavid (approved):

sebrandon1:
Still working through whether or not this is the true fix.  Marked WIP until I can get more test runs logged.

jmontesi:
Wouldn't it be cleaner to use `spec.minReadySeconds` when defining the DaemonSet instead?

PR #477, opened 2023-09-05
Remove unused code for local testing

jmontesi:
The config.yaml file can be overriden with environment variables. There is no need for this special setup.


jmontesi:
Wouldn't it be cleaner to use `spec.minReadySeconds` when defining the DaemonSet instead?

theemj (approved):

PR #479, opened 2023-09-05
Remove automount and projected volume tests

sebrandon1:
Related to: https://github.com/test-network-function/cnf-certification-test/pull/1384

Essentially the pod-service-account test covers the scenarios we need to cover because it forces users to use a service account that is non-default regardless of projected volume, automount enabled/disabled, etc.


sebrandon1:
We had a discussion this morning where we might end up keeping around the projected service account test but just flagging instances where the user might be using a `default` service account token mount.

https://github.com/test-network-function/cnf-certification-test/pull/1385

theemj (approved):

sebrandon1:
Closing in favor of: https://github.com/test-network-function/cnfcert-tests-verification/pull/497

PR #482, opened 2023-09-08
Enable parallel platform-alteration suite

sebrandon1:
image

There are a lot of skipped tests when it comes to Kind clusters in this suite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ gofmt:
 test-all: update-go install-ginkgo download-unstable
 	./scripts/run-tests.sh all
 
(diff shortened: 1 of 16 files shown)


theemj:
Seeing some failures and panics when testing this on the QE cluster, looking into why 
image

sebrandon1:
Let me re-run this.  I think I might want to put some skips in here for `IsKindCluster` for these tests but I'm not sure how that would affect the OCP cluster you're testing on.

sebrandon1:
image

theemj (approved):

PR #491, opened 2023-09-14
DPDK CPU pinning

bnshr:
DPDK pod deployment is based on the cluster setup, existence of HugePages volume, network. Hence, some labels / annotations are hard-coded.


bnshr:
@theemj Could you please test this test case in QE cluster where the cluster setup is ready?

tests/networking/tests/networking_dpdk_cpu_pinning_exec_probe.go
```diff
@@ -0,0 +1,100 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+/*
+	The setup needs to be as mentioned here :
+	https://developers.redhat.com/articles/2021/08/27/using-virtual-functions-dpdk-red-hat-openshift
+*/
+
+var _ = Describe("Networking dpdk-cpu-pinning-exec-probe,", func() {
```
sebrandon1, line 20: Has this worked in a parallel execution?  We can put a `Serial` here if its not ready quite yet and then remove the `Serial` later.

tests/networking/tests/networking_dpdk_cpu_pinning_exec_probe.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+
+	It("one dpdk pod with exec probe [negative]", func() {
+
+		By("Deploy dpdk pod namespace")
+		dpdkPod := tshelper.DefineDpdkPod(tsparams.DpdkPodName, randomNamespace)
+
+		By("Redefine liveness probe")
+
+		containerLivenessProbe := &corev1.Probe{
```
theemj, line 72: We do have a [RedefineWithLivenessProbe function for pods ](https://github.com/test-network-function/cnfcert-tests-verification/blob/4d6bc815c7db3767925a90149e79b9431849547d/tests/utils/pod/pod.go#L52)that could be used here, though it looks like we'd have to add a few params to the function to include all the values being set here. Either way works, might just keep things a little cleaner.

tests/networking/tests/networking_dpdk_cpu_pinning_exec_probe.go
bnshr: Initially I used with Serial when we had one pod deployment in default namespace. But now we moved to pod deployment in random namespace, hence parallel execution would work.

tests/networking/tests/networking_dpdk_cpu_pinning_exec_probe.go
bnshr: Yep. I would modify the function with commands parameter.

theemj (approved):

PR #521, opened 2023-10-03
Add deployment unit tests

sebrandon1:
Adds unit tests for the `deployment.go` file in the globalhelper.

I actually found a potential bug in the `CreateAndWaitUntilDeploymentIsReady` where the scenario where the deployment already exists it should return `nil` in the error but it wasn't and it was sending it into the Eventually loop unnecessarily.

### DIFF
--- a/tests/affiliatedcertification/tests/affiliated_certification_container_is_certified_digest.go
+++ b/tests/affiliatedcertification/tests/affiliated_certification_container_is_certified_digest.go
@@ -45,7 +45,7 @@ var _ = Describe("Affiliated-certification container-is-certified-digest,", func
 		dep := deployment.DefineDeployment("affiliated-cert-deployment", randomNamespace,
 			tsparams.CertifiedContainerURLNodeJs, tsparams.TestDeploymentLabels)
 
-		err := globalhelper.CreateAndWaitUntilDeploymentIsReady(globalhelper.GetAPIClient().K8sClient.AppsV1(), dep, tsparams.Timeout)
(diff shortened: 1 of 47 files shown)


theemj (approved):

tests/affiliatedcertification/tests/affiliated_certification_container_is_certified_digest.go
```diff
@@ -126,13 +126,13 @@ var _ = Describe("Affiliated-certification container-is-certified-digest,", func
 		dep := deployment.DefineDeployment("affiliated-cert-deployment", randomNamespace,
 			tsparams.UncertifiedContainerURLCnfTest, tsparams.TestDeploymentLabels)
 
-		err := globalhelper.CreateAndWaitUntilDeploymentIsReady(dep, tsparams.Timeout)
+		err := globalhelper.CreateAndWaitUntilDeploymentIsReady(globalhelper.GetAPIClient().K8sClient.AppsV1(), dep, tsparams.Timeout)
```
edcdavid, line 129: Would it make sense to create a new helper function instead and make globalhelper.CreateAndWaitUntilDeploymentIsReady a wrapper function just using obalhelper.GetAPIClient().K8sClient.AppsV1(). Also this variable is local to globalhelper. Another bonus is that there would be much less changes/repetitions

sebrandon1: That's a good idea.  I'll do that!

sebrandon1: Done.

edcdavid (approved):

PR #523, opened 2023-10-04
Helm version tests

theemj:


tests/affiliatedcertification/tests/affiliated_certification_helm_version.go
```diff
@@ -0,0 +1,98 @@
...
+			globalparameters.TestCaseFailed)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Re-install helm v3")
+		cmd = exec.Command("/bin/bash", "-c",
+			"curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3"+
+				" && chmod +x get_helm.sh"+
+				" && ./get_helm.sh")
+		err = cmd.Run()
+		Expect(err).ToNot(HaveOccurred(), "Error installing helm v3")
```
sebrandon1, line 96: We should put this in a `DeferCleanup()` func in-case there is a problem with the test and it fails out.

tests/affiliatedcertification/tests/affiliated_certification_helm_version.go
```diff
@@ -0,0 +1,98 @@
...
+	"os/exec"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	tsparams "github.com/test-network-function/cnfcert-tests-verification/tests/affiliatedcertification/parameters"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/globalhelper"
+	"github.com/test-network-function/cnfcert-tests-verification/tests/globalparameters"
+)
+
+var _ = Describe("Affiliated-certification helm-version,", func() {
```
sebrandon1, line 13: We should make this `Serial` as we don't want these two test cases installing and removing helm simultaneous in parallel.

sebrandon1 (approved):

PR #529, opened 2023-10-06
Add _test.go files and script

sebrandon1:
Based on: https://github.com/test-network-function/cnf-certification-test/pull/50


theemj (approved):

scripts/create-missing-test-files.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2043
+for folder in tests; do
```
rdavid, line 3: Why is the cycle needed?

scripts/create-missing-test-files.sh
```diff
@@ -0,0 +1,18 @@
...
+	echo "./$folder"
+
+	FILE_NAMES=$(find ./$folder -name "*.go" | grep -v "_test.go" | grep -v "_moq.go" | grep -v "ginkgo" | sed 's/.go//g')
```
rdavid, line 6:
```suggestion
	FILE_NAMES=$(find ./$folder -name "*.go" ! -name "*_test.go" ! -name "*_moq.go" ! -name "*ginkgo*" | sed 's/.go//g')
```

scripts/create-missing-test-files.sh
```diff
@@ -0,0 +1,18 @@
...
+
+	# if a filename_test.go does not exist, create it
+	for FILE_NAME in $FILE_NAMES; do
+		if [ ! -f "$FILE_NAME"_test.go ]; then
+			touch "$FILE_NAME"_test.go
```
rdavid, line 11: Why is the creation necessary at this stage? The file will be created later.

scripts/create-missing-test-files.sh
rdavid, line 6: I would use lowercase for local variable names.

rdavid (approved):

scripts/create-missing-test-files.sh
sebrandon1: It's a carryover from the cnf-cert-test suite script that has multiple folders with Go code in them.  I'll remove it because all I care about is `/tests`.

scripts/create-missing-test-files.sh
sebrandon1: Removed.

scripts/create-missing-test-files.sh
sebrandon1: Done.

PR #536, opened 2023-10-10
Add assertions for deployments

sebrandon1:
Similar to #534 

This PR adds assertions before running the tests for It() blocks that are using Deployment resources.

MinReadySeconds has also been changed to 15 seconds.

I also added a `NumberReady` check in the IsDeploymentReady func.

### DIFF
--- a/tests/accesscontrol/tests/access_control_no_1337_uid.go
+++ b/tests/accesscontrol/tests/access_control_no_1337_uid.go
@@ -44,6 +44,11 @@ var _ = Describe("Access-control no-1337-uid,", func() {
 		err = globalhelper.CreateAndWaitUntilDeploymentIsReady(dep, tsparams.Timeout)
 		Expect(err).ToNot(HaveOccurred())
 
+		By("Assert deployment has no securityContext RunAsUser 1337")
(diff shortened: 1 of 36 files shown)


tests/accesscontrol/tests/access_control_no_1337_uid.go
```diff
@@ -90,14 +101,24 @@ var _ = Describe("Access-control no-1337-uid,", func() {
...
+		By("Assert deployment has no securityContext RunAsUser")
+		runningDeployment, err := globalhelper.GetRunningDeployment(dep.Namespace, dep.Name)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(runningDeployment.Spec.Template.Spec.SecurityContext.RunAsUser).To(BeNil())
+
 		dep2, err := tshelper.DefineDeployment(1, 1, "accesscontroldeployment2", randomNamespace)
 		Expect(err).ToNot(HaveOccurred())
 
-		deployment.RedefineWithPodSecurityContextRunAsUser(dep, 1338)
+		deployment.RedefineWithPodSecurityContextRunAsUser(dep2, 1338)
```
sebrandon1, line 112: Found a bug here.

PR #594, opened 2023-11-13
Use claim package from the CNF repo

jmontesi:
See [PR#1609](https://github.com/test-network-function/cnf-certification-test/pull/1609) for reference.


rdavid (approved):

sebrandon1 (commented):
I think we need some sort of check or logic added to make sure the version of the test suite that we're including in the go.mod matches the suite checked out on the machine running QE to make sure they are the same version to prevent mismatches.

sebrandon1:
Closing this.  We are going to stick with the -claim repo right now as we are sharing the struct in multiple places and cloning the entire test suite repo gets messy.

PR #608, opened 2023-11-17
Fixed pod's cpu & mem resource setters helpers.

greyerof:
The original implementation was overwriting any req/limit that was previosly set for the other resources. For instance, test cases doing this didn't work properly:
```
  pod.RedefineWithCPUResources(testPod, "1", "1")
  pod.RedefineWithMemoryResources(testPod, "512Mi", "512Mi")
```
The call to pod.RedefineWithMemoryResources() was zero-ing the cpu req/limits that was just set before ("1", "1") as it was creating a new empty container.Resources field (corev1.ResourceRequirements{}).

The new implementations don't overwrite other reqs/limits values, they just set/overwrite the corresponding ones (cpu/mem).

The affected test cases were passing until now because a bug in the cert suite code in the function AreCPUResourcesWholeUnits() that was wrongly returning "true" when the cpu reqs/limit were not set. It was fixed here:
https://github.com/test-network-function/cnf-certification-test/pull/1631

### DIFF
--- a/tests/utils/pod/pod.go
+++ b/tests/utils/pod/pod.go
@@ -102,27 +102,35 @@ func RedefineWithPVC(pod *corev1.Pod, volumeName string, claimName string) {
 
 func RedefineWithCPUResources(pod *corev1.Pod, limit string, req string) {
 	for i := range pod.Spec.Containers {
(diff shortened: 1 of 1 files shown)


greyerof:
@sebrandon1 this should fix PR #607

sebrandon1 (approved):

PR #616, opened 2023-11-30
Switch to 1-to-1 result struct mapping

sebrandon1:
Related to: 
- https://github.com/test-network-function/cnf-certification-test/pull/1675
- https://github.com/test-network-function/cnf-certification-test/pull/1666

Depends-On: https://github.com/test-network-function/cnf-certification-test/pull/1675


sebrandon1:
I merged this because the 'Depends-On' action wasn't doing what I wanted it to do.

PR #647, opened 2023-12-22
Update claim repo dependency to v1.0.33

greyerof:


greyerof:
The failures in QE suites are normal due to the circular dependency with the CNF Cert Suite repo. I'll re-run them once https://github.com/test-network-function/cnf-certification-test/pull/1762 is merged

greyerof:
The jobs passed when re-launched manually after the PR https://github.com/test-network-function/cnf-certification-test/pull/1762 was merged.
See https://github.com/test-network-function/cnfcert-tests-verification/actions/runs/7298934998

PR #675, opened 2024-01-18
Remove free disk space cleanup

sebrandon1:
Similar to: https://github.com/test-network-function/cnf-certification-test/pull/1821

### DIFF
--- a/.github/workflows/qe.yml
+++ b/.github/workflows/qe.yml
@@ -29,18 +29,6 @@ jobs:
       SKIP_PRELOAD_IMAGES: true # Not needed for github-hosted runs
 
     steps:
-      - name: Free Disk Space (Ubuntu)
-        uses: jlumbroso/free-disk-space@main
-        with:
-          tool-cache: false
-          large-packages: true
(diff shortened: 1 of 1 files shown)


.github/workflows/qe.yml
```diff
@@ -29,17 +29,17 @@ jobs:
...
-        with:
-          tool-cache: false
-          large-packages: true
-          android: true
-          dotnet: true
-          haskell: true
-          docker-images: true
-          swap-storage: true
-        continue-on-error: true
+      # - name: Free Disk Space (Ubuntu)
```
rdavid, line 32: We have a Git history; there is no need for commented-out code.

rdavid (approved):

.github/workflows/qe.yml
sebrandon1: I'll remove it once it actually works I think.  I'm not convinced the changes are fully rolled out yet.

sebrandon1: Done.

PR #686, opened 2024-02-08
Update GolangCI-lint to v1.56.0

sebrandon1:
https://github.com/golangci/golangci-lint/releases/tag/v1.56.0


rdavid (approved):

sebrandon1:
😄 This is already out of date.  They always do this.  I'm going to wait a couple of weeks before bumping this I think.

PR #696, opened 2024-02-19
Shorten minReadySeconds

sebrandon1:
Changing the `minReadySeconds` from 15 seconds to 5 seconds across the repo.  Wondering if all of the stability changes will allow for this work happily.


sebrandon1:
Comparing against an [older](https://github.com/test-network-function/cnfcert-tests-verification/actions/runs/7953111121) nightly run, this is ~4 minutes faster to run with each suite individually running faster.

PR #703, opened 2024-03-06
Do not expect error on negative test cases

jmontesi:
The behavior of the `cnf-certification-test` binary has been modified to return an error code only in case of an actual runtime error and not due to failed test cases.

Depends-On: [test-network-function/cnf-certification-test#1901](https://github.com/test-network-function/cnf-certification-test/pull/1901)


sebrandon1:
This looks good.  Because we use the `unstable` image I don't think the depends-on keyword will work here but I understand the change.

sebrandon1 (approved):

PR #707, opened 2024-03-15
Switch expected output for no labels

sebrandon1:
Follow up to #706 with another test case that is now expected failed due to: https://github.com/test-network-function/cnf-certification-test/pull/1871


sebrandon1:
image

PR #778, opened 2024-05-23
Remove afterEach cleanup for re-enabled tests

sebrandon1:
Bonus: Fully enabled parallel processing for the `operator` suite.


sebrandon1:
Marking as WIP as there are still some tests that are not happy in OCP.

PR #784, opened 2024-05-28
Build the binary and test against

sebrandon1:
Related to work done: https://github.com/test-network-function/cnf-certification-test/pull/2095

### DIFF
--- a/tests/observability/tests/crd_status.go
+++ b/tests/observability/tests/crd_status.go
@@ -23,6 +23,24 @@ const (
 )
 
 var _ = Describe(tsparams.TnfCrdStatusTcName, Serial, func() {
+	var randomNamespace string
+	var randomReportDir string
+	var randomTnfConfigDir string
+
+	BeforeEach(func() {
+		// Create random namespace and keep original report and TNF config directories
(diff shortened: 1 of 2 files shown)


tests/observability/tests/crd_status.go
```diff
@@ -47,13 +67,13 @@ var _ = Describe(tsparams.TnfCrdStatusTcName, Serial, func() {
 		By("Start TNF " + tsparams.TnfCrdStatusTcName + " test case")
 		err = globalhelper.LaunchTests(tsparams.TnfCrdStatusTcName,
 			globalhelper.ConvertSpecNameToFileName(CurrentSpecReport().FullText()),
-			globalhelper.GetConfiguration().General.TnfReportDir,
```
rdavid, line 50: Is `globalhelper.GetConfiguration().General` used by anyone? If not, it should be removed.

rdavid (approved):

tests/observability/tests/crd_status.go
sebrandon1: Good question, maybe another PR to address that one.

PR #785, opened 2024-05-29
Revert ubuntu-24.04 change due to cancelled runs

sebrandon1:


rdavid (approved):

sebrandon1:
Seems others are also seeing the cancellation failure: https://github.com/actions/runner-images/issues/9848#issuecomment-2137525256

PR #803, opened 2024-06-13
Fix prefix for openvino operator

bnshr:


aabughosh (approved):

sebrandon1:
I don't think this fixes it completely.  This is just for the error message printing.

bnshr:
> I don't think this fixes it completely. This is just for the error message printing.

Interesting. I ran them in 3 nodes cluster and worked fine.

PR #921, opened 2024-09-13
Add clusterrole pkg from eco-goinfra

sebrandon1:
Similar to: #895


sebrandon1:
Picked up changes from: https://github.com/openshift-kni/eco-goinfra/pull/659 now that it has been merged.

PR #935, opened 2024-09-30
observability: add more cases to pod-disruption-budget

jmontesi:
Four new test cases are added to verify the proper matching of a PDB's matchLabels and matchExpresssions to a Deployment's labels.


sebrandon1 (approved):
Nice!  I like the new `maps.Clone` usage from Go 1.21.

PR #973, opened 2024-10-29
operator: add test for multiple-same-operators

sebrandon1:
Adding tests for the new multiple-same-operators test case.

This PR includes a positive and negative test.

The positive test involves installing the same operator.version in two different namespaces.  This is a valid case.

The negative test involves installing different operator.version's in two different namespaces.  This is an invalid case.


rdavid (approved):

sebrandon1:
I would say any `operator` or `affiliatedcertification` test cases in the future should probably setup and tear-down their desired catalogsources in the future.  Or at least, verifying that they are there prior to running the tests.

PR #1000, opened 2024-11-26
Delete istio CRDs when uninstalling istio.

greyerof:
This is an attemp to fix the flaky tc of the affiliatedcertification test suite that has been failing too often lately.

Istio CRDs must be removed when uninstalling istio. Otherwise, this error might appear when re-installing it:

"If istio is not uninstalled properly, the next installation retry fails because of this error:
Error: INSTALLATION FAILED: Unable to continue with install: CustomResourceDefinition "wasmplugins.extensions.istio.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal
"affiliatedcert-tests-hrzzgmvrkr": current value is "affiliatedcert-tests-maahivhcff"

See https://github.com/istio/istio/issues/43204

Also:
- DeferCleanup func installed before the step that installs istio.
- cmd.Run() changed to cmd.CombinedOutput() so cmd output can be shown in the gomega error assertion message. If this fix doesn't work, at least we'll get more info from the commands that install/uninstall istio.
- Added some error return wrappings and minor code refactors.
- Avoided abrupt exit in some scenarios to allow the normal recovery to continue.
- Avoid the need of certsuite repo env var if running in container mode.

### DIFF
--- a/tests/utils/crd/crd.go
+++ b/tests/utils/crd/crd.go
@@ -143,7 +143,7 @@ func GetAllCustomResourceDefinitions() ([]apiextv1.CustomResourceDefinition, err
 
 func DeleteCustomResourceDefinition(name string) error {
 	err := globalhelper.GetAPIClient().CustomResourceDefinitions().Delete(context.TODO(), name, metav1.DeleteOptions{})
-	if err != nil {
+	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
 	}
 


tests/utils/crd/crd.go
```diff
@@ -131,3 +131,21 @@ func CreateCustomResourceScale(name, namespace, operatorLabels string, operatorL
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list CRDs: %w", err)
+	}
+
+	return crdList.Items, nil
+}
+
+func DeleteCustomResourceDefinition(name string) error {
+	err := globalhelper.GetAPIClient().CustomResourceDefinitions().Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if err != nil {
```
sebrandon1, line 146: Should this also handle `IfNotExist` scenarios when the CRD is already gone?

PR #1012, opened 2024-12-03
Adjust typo

sebrandon1:
This looks like a typo from when we changed over to the new org.


sebrandon1:
This might actually have to be the working dir of the github runner.

rdavid (approved):

PR #1032, opened 2024-12-19
Adjusted operator-install test case.

greyerof:
Certsuite's PR
https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2653 changed the way CSVs are autodiscovered in order for operator pods to be tested: the operator installation namespace must be explicitly added to the configured namespaces.

This change deploys the operatorgroup and subscription of the logging operator in the random test namespace that is set in the certsuite config file.


greyerof:
Checks passed. Merging it so we can merge certsuite's [PR2653](https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2653) with the main qe ref switched back to main.

PR #1034, opened 2024-12-20
Rename access control test

greyerof:
Due to this certsuite PR:
https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2654 ,
the runAsUser and runAsNonRoot securityContext fields check has been
merged into access-control-security-context-run-as-non-root-user-check.


greyerof:
I'll rework the certsuite PR to use the other tc name so we don't need to change this repo.

PR #1063, opened 2025-02-10
Add log output for debugging when helm installation fails.

greyerof:
Just added the stdout+stderr output of the helm tool installation as the affiliated test suite seems to fail too often. As this happens at the SynchronizedBeforeSuite() func, all the test cases are skipped when it happens.

I've seen this failure in all our OCP versions where we run QE.

Last failing runs in 4.16:
- https://github.com/redhat-best-practices-for-k8s/certsuite/actions/runs/12919752505
- https://github.com/redhat-best-practices-for-k8s/certsuite/actions/runs/13001158158
- https://github.com/redhat-best-practices-for-k8s/certsuite/actions/runs/13220699737


sebrandon1:
So I think I know why it fails.  It's probably a concurrency problem where there are other runs happening on the same jump box (the same machine runs all 4 runner scripts) and its entirely possible that it's trying to install/remove helm between the different runs at the same time.  Let's merge this so we have better insight.

sebrandon1 (approved):

PR #1442, opened 2026-04-17
Fix performance scheduling tests and add precondition assertions

sebrandon1:
## Summary
- Switches `SampleWorkloadImage` for performance tests to `quay.io/redhat-best-practices-for-k8s/certsuite-sample-workload:latest` (which includes `chrt` via `util-linux`), fixing the `chrt -p 1` exit code 255 failure that has broken all nightly runs since April 10
- Adds certsuite-aligned precondition assertions to all three `shared-cpu-pool-non-rt-scheduling-policy` test cases so failures due to incorrect environment setup are caught early with clear messages

## Depends on
- redhat-best-practices-for-k8s/certsuite-sample-workload#677 (adds `util-linux` to the sample workload image)

### Test 1 — shared CPU pool
- Asserts pod QoS is not Guaranteed (certsuite skips guaranteed pods)
- Asserts pod does not use HostPID (certsuite skips HostPID pods)

### Test 2 — ephemeral subprocesses
- Same QoS and HostPID assertions
- Verifies subprocesses are actually running via `pgrep` before launching certsuite

### Test 3 — exclusive CPU pool
- Asserts pod QoS is Guaranteed
- Asserts pod does not use HostPID
- Retains CPU whole-unit and limits=requests assertions

## Test plan
- [ ] Verify linter passes (confirmed locally)
- [ ] Verify performance tests pass on OCP 4.14/4.16/4.18 after sample workload image is updated

### DIFF
--- a/tests/performance/parameters/parameters.go
+++ b/tests/performance/parameters/parameters.go
@@ -43,5 +43,5 @@ const (
 
 	DisableStr = "disable"
 
-	SampleWorkloadImage = "registry.access.redhat.com/ubi8/ubi-micro:latest"
+	SampleWorkloadImage = "quay.io/redhat-best-practices-for-k8s/certsuite-sample-workload:latest"
 )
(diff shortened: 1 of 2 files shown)


sebrandon1:
Reopening from upstream branch to trigger OCP tests.

